PR #2, opened 2017-11-28
Expand tabs in Include/patchlevel.h.

serhiy-storchaka:


ned-deily:
LGTM.  Thanks, Serhiy!

PR #5, opened 2020-11-09
Simplify Windows download entries

zooba:
The main aim of this is to reduce the number of items shown for Windows items, as well as to bring them more in line with how people are used to seeing them. Specifically:

* remove entries for pre-3.5 releases from the script completely
* change "x86" and "x86-64" to "(32-bit)" and "(64-bit)" (which matches the installed Start menu entries)
* remove the webinstall entries (not widely used)

There doesn't seem to be any easy way to make these changes not apply to 3.8 and 3.9, but I don't really mind if those are updated as well.

I also added the `print_function` future and put parentheses everywhere so I didn't have to install Python 2 just to double-check my syntax.

### DIFF
--- a/add-to-pydotorg.py
+++ b/add-to-pydotorg.py
@@ -22,6 +22,7 @@ To use (RELEASE is something like 3.3.5rc2):
 
 Georg Brandl, March 2014.
 """
+from __future__ import print_function
 
 import hashlib
 import json
@@ -36,8 +37,8 @@ import requests
 try:
     auth_info = os.environ['AUTH_INFO']
 except KeyError:
-    print 'Please set an environment variable named AUTH_INFO ' \
-        'containing "username:api_key".'
+    print('Please set an environment variable named AUTH_INFO '
+        'containing "username:api_key".')
     sys.exit()
 
 base_url = 'https://www.python.org/api/v1/'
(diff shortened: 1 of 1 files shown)


zooba:
Ping @ambv @pablogsal

add-to-pydotorg.py
```diff
@@ -22,6 +22,7 @@
 
 Georg Brandl, March 2014.
 """
+from __future__ import print_function
```
pablogsal, line 25: Oh boy, I think I would never see you again in this GitHub org :P

pablogsal (approved):
LGTM

add-to-pydotorg.py
ned-deily: It looks like the download server, where RMs run the script, currently does have an Ubuntu 3.6.9 installed, so we could consider making the big leap to running under Python 3 :)

ned-deily (approved):

add-to-pydotorg.py
zooba: I could have left it as only valid syntax for Python 2.x if you liked :)

zooba:
Just want to make sure @ambv sees this before it goes in, in case it becomes a surprise for the next 3.8/3.9 releases.

PR #11, opened 2021-10-03
Make the macOS universal2 installer variant the default download.

ned-deily:
For Python 3.10, we only manufacture the universal2 variant for macOS.

For Python 3.9.x, we produce universal2 and legacy Intel-only
With this change, the next release of 3.9.x, presumably 3.9.8,
will now download the universal2 variant by default rather than
the Intel-only one; we have been warning of this anticipated
change in recent 3.9.x release announcements and in the
macOS installer Read Me splash screen and file.

Release managers will need to update their copies of add-to-pytdotorg.


pablogsal:
Updated on my side!

PR #14, opened 2022-05-09
Dynamicize previous version in whatsnew template.

ezio-melotti:
Should prevent https://github.com/python/cpython/pull/92488 from happening again.
The PR is untested.


pablogsal (approved):

pablogsal:
Thanks a lot @ezio-melotti ❤️

PR #15, opened 2022-05-17
Import Azure Pipelines build script for Windows releases

zooba:
This is the script that's currently in the main repo, but doesn't really need to be there.


zooba:
My test build has passed, so I'm merging it in.

PR #17, opened 2022-08-11
Sign w/ Sigstore and upload Sigstore verification materials

di:
Pairs with https://github.com/python/pythondotorg/pull/2113.

I didn't see where non-source release artifacts were getting signed, looks like it's happening outside this process, but if we want to include that here too, point me in the right direction and I'll add it.

(cc @python/python-release-managers)

### DIFF
--- a/add-to-pydotorg.py
+++ b/add-to-pydotorg.py
@@ -144,7 +144,7 @@ def list_files(release):
     for rfile in os.listdir(path.join(ftp_root, reldir)):
         if not path.isfile(path.join(ftp_root, reldir, rfile)):
             continue
-        if rfile.endswith('.asc'):
+        if rfile.endswith(('.asc', '.sig', '.crt')):
             continue
         for prefix in ('python', 'Python'):
             if rfile.startswith(prefix):
--- a/add-to-pydotorg.py
+++ b/add-to-pydotorg.py
@@ -34,6 +34,8 @@ import time
 
 import requests
 
+from release import run_cmd
+
 try:
     auth_info = os.environ['AUTH_INFO']
 except KeyError:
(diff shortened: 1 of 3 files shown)


pablogsal:
> I didn't see where non-source release artifacts were getting signed, looks like it's happening outside this process, but if we want to include that here too, point me in the right direction and I'll add it.

The script is executed directly on the download server, where all the artefacts are already stored. The signing is done before and all the .asc files are uploaded to the server before running the script by the platform release managers

di:
Got it. Since we ultimately want the release manager's signature on these, should we modify the script to just sign everything w/ Sigstore, not just the source distributions?

pablogsal:
> should we modify the script to just sign everything w/ Sigstore, not just the source distributions?

Yeah, that sounds good 👍 

Btw, IIRC, doesn't the signing process open a browser? That won't work on the server. Can it print the URL instead?

di:
> Btw, IIRC, doesn't the signing process open a browser? That won't work on the server. Can it print the URL instead?

The Sigstore client will autodetect if a browser is available and will print a URL if one is not.

> Yeah, that sounds good 👍

Ok, I'll work on this.

di:
@pablogsal I think this should be good to go, but it's hard for me to test.

This signs the source releases immediately _after_ they're made, and signs everything else immediately _prior_ to upload.

add-to-pydotorg.py
```diff
@@ -34,6 +34,8 @@
 
 import requests
 
+from release import run_cmd
```
pablogsal, line 37: Hummm, this needs to be copied here because this script live standalone in the server. The `release.py` file will not be available there.

pablogsal: Checkout https://github.com/python/release-tools/blob/3562448d6042a48d3bf19c7c808654c2137fd4a0/run_release.py#L720-L744

pablogsal: Apologies for the trouble 😅

di: I had a suspicion this was the case, but missed that. Should be good now!

pablogsal:
LGTM

Thanks for the changes @di

PR #18, opened 2022-08-18
Add Windows OpenSSL build scripts

zooba:


zooba:
See https://github.com/python/cpython/pull/96088

PR #22, opened 2022-10-05
Fix issues with Sigstore signing

### LINKED ISSUE #21: Issues w/ Sigstore signing
Hey @pablogsal, sorry for the troubles. Just want to make sure I understand all the issues you hit. To summarize https://github.com/python/pythondotorg/pull/2113 and  https://github.com/python/release-tools/pull/19, what we'll need is:

- [ ] file permissions are wrong, this needs updated in <code class="notranslate">add_to_python_dot_org.py</code> to do it directly on the host.
- [ ] Automation is broken because interactive signing happens in <code class="notranslate">add-to-pydotorg.py</code>, but this is run non-interactively
- [x] NameError (resolved already by https://github.com/python/release-tools/pull/19)
- [ ] revert https://github.com/python/release-tools/pull/20

Is that right?

Regarding the blocking for signing, this is expected, but should happen in an interactive environment, so looks like `add_to_python_dot_org.py` is the wrong place for this. Is there any point in the release process where the files are available in an interactive environment?

di:
Fixes #21.

Note that this requires `sigstore` to be installed on both the local machine (to generate an identity token) and the remote machine (to actually do the signing). This also requires the newly released `sigstore>0.6.6`, I've updated the `requirements.txt` file to match.

I _think_ this will correctly pass `IDENTITY_TOKEN` through all the right layers to be present when Sigstore is invoked on the remote machine, but I wasn't able to test this.

### DIFF
--- a/add-to-pydotorg.py
+++ b/add-to-pydotorg.py
@@ -155,10 +155,10 @@ def build_file_dict(release, rfile, rel_pk, file_desc, os_pk, add_desc):
         d["gpg_signature_file"] = sigfile_for(base_version(release), rfile)
     # Upload Sigstore signature
     if os.path.exists(ftp_root + "%s/%s.sig" % (base_version(release), rfile)):
-        d["sigstore_signature_file"] = download_root + '%s/%s.sig' % (release, rfile)
+        d["sigstore_signature_file"] = download_root + '%s/%s.sig' % (base_version(release), rfile)
     # Upload Sigstore certificate
     if os.path.exists(ftp_root + "%s/%s.crt" % (base_version(release), rfile)):
-        d["sigstore_cert_file"] = download_root + '%s/%s.crt' % (release, rfile)
+        d["sigstore_cert_file"] = download_root + '%s/%s.crt' % (base_version(release), rfile)
 
     return d
 
--- a/add-to-pydotorg.py
+++ b/add-to-pydotorg.py
@@ -244,6 +244,10 @@ def sign_release_files_with_sigstore(release, release_files):
     print('Signging release files with Sigstore')
     run_cmd(['python3', '-m', 'sigstore', 'sign', '--oidc-disable-ambient-providers'] + filenames)
 
+    # Update permissions on Sigstore verification materials to match other downloads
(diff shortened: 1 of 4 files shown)


release.py
```diff
@@ -271,9 +271,8 @@ def tarball(source):
     os.system('gpg -bas -u ' + uid + ' ' + tgz)
     os.system('gpg -bas -u ' + uid + ' ' + xz)
 
-    # TODO: Uncomment when this is synced with the rest of the release scripts
-    # print('Signging tarballs with Sigstore')
-    # run_cmd(['python3', '-m', 'sigstore', 'sign', '--oidc-disable-ambient-providers', tgz, xz])
+    print('Signging tarballs with Sigstore')
```
ezio-melotti, line 274:
```suggestion
    print('Signing tarballs with Sigstore')
```

run_release.py
```diff
@@ -732,8 +732,10 @@ def run_add_to_python_dot_org(db: DbfilenameShelf) -> None:
 
     auth_info = db["auth_info"]
     assert auth_info is not None
+    # Do the interactive flow to get an identity for Sigstore
+    identity_token = subprocess.check_output(["python", "-m", "sigstore", "get-identity-token"]).strip().decode()
```
pablogsal, line 736: Apparently the output of this line includes "Waiting for browser interaction...". We probably need to remove that

add-to-pydotorg.py
```diff
@@ -244,14 +244,17 @@ def has_sigstore_signature(rfile):
     print('Signging release files with Sigstore')
     run_cmd(['python3', '-m', 'sigstore', 'sign', '--oidc-disable-ambient-providers'] + filenames)
 
+    # Update permissions on Sigstore verification materials to match other downloads
+    run_cmd(["find", ftp_root, "-type", "f", "\(", "-iname", "\*.sig", "-o", "-iname", "\*.crt", "\)", "exec", "chmod", "644", "{}", "\\;"])
```
ambv, line 248: I reverted this in 21e1adaf9b04ffbcf8d3699bf11f1ecac3db239c because I'm not okay with touching all files in ftp_root. Ideally `sigstore sign` should produce properly accessible .sig and .crt files. But in the absence of this, chmodding *only* the files from `unsigned_files` (+ .sig/.crt extensions) would be okay.

ambv: I made this suggested change in 6846215910b2bd6db82f21fd68289afe8c5b8f76.

PR #53, opened 2023-07-24
Use a lockfile with hashes for dependencies

sethmlarson:
Currently the list of exact dependency versions in use isn't known by contributors to release-tools. Using a lock file means there aren't surprises/regressions when release managers go to make a release. Also has the benefit of Dependabot alerting us of vulnerabilities in dependencies in use (I believe it will work even if not explicitly enabled in the UI)


ambv:
Thanks! ✨ 🍰 ✨

PR #61, opened 2023-10-04
Check commit SHA after checking out repository

sethmlarson:
This adds an optional parameter and step to the "checkout" process which verifies that a given git tag resolves to a known commit SHA. This makes it far more difficult for someone with access to the release managers' fork of CPython on GitHub to spoof the inputs to the Windows installer builds (for example, changing the tag to a separate commit, and then changing it back after the build has started).

This requires specifying the git commit that's expected, so that value should be shared by the release manager through a separate channel to ensure the right value is used.

After this PR is accepted, I can propose an update of PEP 101 which provides this additional instruction for creating release builds.

My PowerShell is not up to par, I haven't ran Windows in around 8 years so please check that it's correct :)

### DIFF
--- a/windows-release/checkout.yml
+++ b/windows-release/checkout.yml
@@ -23,7 +23,7 @@ steps:
 - powershell: |
     $checkout_commit = (git rev-parse HEAD)
     if ($checkout_commit -ne '$(SourceCommit)') {
-        throw "Expected git commit '$(SourceCommit)' didn't match tagged commit '$(checkout_commit)'"
+        throw "Expected git commit '$(SourceCommit)' didn't match tagged commit '$checkout_commit'"
     }
   displayName: "Verify CPython commit matches tag"
   condition: and(succeeded(), variables['SourceCommit'])
--- a/windows-release/checkout.yml
+++ b/windows-release/checkout.yml
@@ -19,3 +19,11 @@ steps:
 - script: git clone --progress -v --depth ${{ parameters.depth }} --branch $(Build.SourceBranchName) --single-branch $(Build.Repository.Uri) .
   displayName: 'git clone'
   condition: and(succeeded(), and(not(variables['GitRemote']), not(variables['SourceTag'])))
+
+- powershell: |
+    $checkout_commit = (git rev-parse HEAD)
+    if ($checkout_commit -ne $GitCommit) {
+        throw "Expected git commit '$(GitCommit)' didn't match tagged commit '$(checkout_commit)'"
+    }
+  displayName: "Verify CPython commit matches tag"
+  condition: and(succeeded(), and(variables['GitCommit']))
--- a/windows-release/checkout.yml
+++ b/windows-release/checkout.yml
@@ -22,8 +22,8 @@ steps:
 
 - powershell: |
     $checkout_commit = (git rev-parse HEAD)
-    if ($checkout_commit -ne $GitCommit) {
-        throw "Expected git commit '$(GitCommit)' didn't match tagged commit '$(checkout_commit)'"
+    if ($checkout_commit -ne SourceCommit) {
+        throw "Expected git commit '$(SourceCommit)' didn't match tagged commit '$(checkout_commit)'"
     }
   displayName: "Verify CPython commit matches tag"
-  condition: and(succeeded(), and(variables['GitCommit']))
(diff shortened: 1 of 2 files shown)


zooba:
Looks fine to me. Don't suppose you're able to push the branch directly to this repo? That way I can run a test build easily (i.e. without modifying the build definition to link to your GitHub account)

sethmlarson:
@zooba Done, pushed to `windows-check-commit-sha`

zooba:
Well here's problem number 1 😆 

image

The way I usually get around this is to have a default value of "skip" or similar, and then check for that. An empty parameter gets set to the default, and the default can't be empty, which means you end up making it required.

sethmlarson:
@zooba Oh huh, so having a `default` of `''` isn't making the parameter optional! That's unfortunate, I'll push a sentinel value there then.

windows-release/checkout.yml
```diff
@@ -19,3 +19,11 @@ steps:
...
   displayName: 'git clone'
   condition: and(succeeded(), and(not(variables['GitRemote']), not(variables['SourceTag'])))
+
+- powershell: |
+    $checkout_commit = (git rev-parse HEAD)
+    if ($checkout_commit -ne $GitCommit) {
+        throw "Expected git commit '$(GitCommit)' didn't match tagged commit '$(checkout_commit)'"
+    }
+  displayName: "Verify CPython commit matches tag"
+  condition: and(succeeded(), and(variables['GitCommit']))
```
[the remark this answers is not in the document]
sethmlarson: Ah, this was likely a copy-paste error. I'll fix this up.

sethmlarson:
@zooba I pushed a commit to both remotes, you can test again :)

zooba:
I ran three test builds:
* [should skip the check](https://dev.azure.com/Python/cpython/_build/results?buildId=139762&view=results)
* [should fail the check](https://dev.azure.com/Python/cpython/_build/results?buildId=139763&view=results)
* [should pass the check](https://dev.azure.com/Python/cpython/_build/results?buildId=139764&view=results)

windows-release/checkout.yml
```diff
@@ -19,3 +19,11 @@ steps:
 - script: git clone --progress -v --depth ${{ parameters.depth }} --branch $(Build.SourceBranchName) --single-branch $(Build.Repository.Uri) .
   displayName: 'git clone'
   condition: and(succeeded(), and(not(variables['GitRemote']), not(variables['SourceTag'])))
+
+- powershell: |
+    $checkout_commit = (git rev-parse HEAD)
+    if ($checkout_commit -ne SourceCommit) {
```
zooba, line 25:
```suggestion
    if ($checkout_commit -ne '$(SourceCommit)') {
```

This should be fine, though injecting pipeline variables into Powershell can be a bit flakey...

sethmlarson:
Updated, thanks for the diff.

sethmlarson:
Had to fix the author, apparently `sethmichaellarson@gmail.com` hasn't signed the CLA :)

windows-release/checkout.yml
```diff
@@ -19,3 +19,11 @@ steps:
 - script: git clone --progress -v --depth ${{ parameters.depth }} --branch $(Build.SourceBranchName) --single-branch $(Build.Repository.Uri) .
   displayName: 'git clone'
   condition: and(succeeded(), and(not(variables['GitRemote']), not(variables['SourceTag'])))
+
+- powershell: |
+    $checkout_commit = (git rev-parse HEAD)
+    if ($checkout_commit -ne '$(SourceCommit)') {
+        throw "Expected git commit '$(SourceCommit)' didn't match tagged commit '$(checkout_commit)'"
```
zooba, line 26:
```suggestion
        throw "Expected git commit '$(SourceCommit)' didn't match tagged commit '$checkout_commit'"
```

This should be the last fix. I think we've got everything now

sethmlarson: Fixed and pushed! :)

zooba:
That's looking good! https://dev.azure.com/Python/cpython/_build/results?buildId=139771&view=logs&j=9853e208-2957-52e0-8166-2429e0ccf669&t=4f74a7e5-69e9-5d77-9e0b-a4d9e6fbb343&l=21

And my other two builds have successfully skipped/passed the check, so I think we're good to merge.

PR #63, opened 2023-10-06
Now pulls redistributable VC files from cpython-bin-deps

zooba:
This ensures we will not downgrade DLL versions between releases, which leads to issues such as https://github.com/python/cpython/issues/110437


zooba:
Test build running at https://dev.azure.com/Python/cpython/_build/results?buildId=139956&view=results

PR #75, opened 2023-11-15
Use manifest for quay.io/tiran/cpython_autoconf instead of tag

### LINKED ISSUE #70: Pin the cpython_autoconf container to a sha256 manifest
The container hasn't needed an update in 2 years. I propose pinning the container image used by the release process to the sha256 manifest instead of the tag in order to prevent any compromise there from being automatically used during the release.

sethmlarson:
Closes #70 This would prevent any pushes to the autoconf container from immediately affecting the source builds. I also removed the `check_autoconf` job since I don't think it's necessary anymore w/ the Dockerized autoconf?

### DIFF
--- a/run_release.py
+++ b/run_release.py
@@ -298,7 +298,6 @@ check_git = functools.partial(check_tool, tool="git")
 check_latexmk = functools.partial(check_tool, tool="latexmk")
 check_make = functools.partial(check_tool, tool="make")
 check_blurb = functools.partial(check_tool, tool="blurb")
-check_autoconf = functools.partial(check_tool, tool="autoconf")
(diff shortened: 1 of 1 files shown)


zware:
Note that the build was *just* changed to avoid the docker container due to instability in `quay.io`.  See python/cpython#112090, commit python/cpython@d9fd33a869d2be769ff596530f63ee099465b037.

sethmlarson:
@zware Thanks for the note, that makes sense! Glad that Ubuntu is providing a proper autoconf now. Maybe I'll move the hardening work there and swap the release scripts to use that file when available?

sethmlarson:
Closing this in favor of https://github.com/python/cpython/issues/112160

PR #79, opened 2023-12-15
Use 'regen-configure' Makefile target

sethmlarson:
Follow-up from https://github.com/python/cpython/issues/112160
Closes https://github.com/python/release-tools/issues/70

This uses `make regen-configure` instead of pulling in the Docker container. In the latest CPython branch we no longer use the autoconf container and in past branches the container is pinned to a hash which is preferable over a tag.

### DIFF
--- a/run_release.py
+++ b/run_release.py
@@ -412,17 +412,8 @@ def prepare_pydoc_topics(db: DbfilenameShelf) -> None:
 
 def run_autoconf(db: DbfilenameShelf) -> None:
     subprocess.check_call(
-        [
-            "docker",
-            "run",
-            "--rm",
-            "--pull=always",
-            f"-v{db['git_repo']}:/src",
-            "quay.io/tiran/cpython_autoconf:cp311",
-        ],
-        cwd=db["git_repo"],
+        ["make", "regen-configure"], cwd=db["git_repo"],
     )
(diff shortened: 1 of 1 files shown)


ezio-melotti (approved):
LGTM.  See also:
* https://github.com/python/devguide/pull/1236

sethmlarson:
Nice, thanks @ezio-melotti for the review and link!

Yhg1s:
This does not seem like the correct change:
```
✅  Preparing pydoc topics
make: *** No rule to make target 'regen-configure'.  Stop.
💥  Running autoconf
Traceback (most recent call last):
  File "./run_release.py", line 1058, in <module>
    main()
  File "./run_release.py", line 1054, in main
    automata.run()
  File "./run_release.py", line 253, in run
    raise e from None
  File "./run_release.py", line 250, in run
    self.current_task(self.db)
  File "./run_release.py", line 188, in __call__
    return getattr(self, "function")(db)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./run_release.py", line 414, in run_autoconf
    subprocess.check_call(
  File "/usr/lib/python3.11/subprocess.py", line 413, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['make', 'regen-configure']' returned non-zero exit status 2.
```

The point at which configure is regenerated is before configure is run, so there is no Makefile and thus no 'regen-configure' make target.

PR #81, opened 2023-12-20
Implement free-threaded build for Windows release

zooba:


zooba:
Test build running at https://dev.azure.com/Python/cpython/_build/results?buildId=145857&view=results (which I fully expect to fail somewhere, I'm just not sure where)

zooba:
Just ran two more test builds with my main patch and the option on and off, so it all looks good.

PR #84, opened 2024-01-17
Add potential SBOMs to python.org

sethmlarson:
Part of https://github.com/python/cpython/issues/112302

Currently this code has no effect, however once SBOMs begin to be generated in https://github.com/python/release-tools/pull/82 there will be additional files so we need to do something with them to avoid failing releases on "unexpected file"

This PR is a follow-up from https://github.com/python/pythondotorg/pull/2359, so python.org now supports the field `sbom_spdx2_file`.


hugovk (approved):
Looks good.

I was going to suggest some f-strings, but this is consistent with existing code.

We might benefit from adding some CI (in another PR), even if just some linting, it could help catch potential syntax errors and so on.

sethmlarson:
@hugovk Agreed, I created a shell issue for myself: https://github.com/python/release-tools/issues/85

Issue #83, opened 2024-01-17
Issue with make regen-configure


sethmlarson:
See: https://github.com/python/release-tools/pull/79#issuecomment-1893953566

PR #87, opened 2024-01-18
Run regen-configure script if available

sethmlarson:
Closes https://github.com/python/release-tools/issues/83 @Yhg1s was this the fix you were imagining?

### DIFF
--- a/run_release.py
+++ b/run_release.py
@@ -411,9 +411,30 @@ def prepare_pydoc_topics(db: DbfilenameShelf) -> None:
 
 
 def run_autoconf(db: DbfilenameShelf) -> None:
-    subprocess.check_call(
-        ["make", "regen-configure"], cwd=db["git_repo"],
-    )
+    # Python 3.12 and newer have a script that runs autoconf.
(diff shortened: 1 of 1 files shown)


Yhg1s:
I guess this could work, but it's not what I had in mind, no. I had in mind just running the steps needed to regen configure: 

```
     subprocess.check_call(
        [db["git_repo"] / "Tools/build/regen-configure.sh"],
     )
```

... but that will only work for 3.12 and later. 3.10 and 3.11 still require the use of podman/docker and Christian's images. We could backport the regen-configure script to those releases and make it then run docker the way the makefile currently does, I guess. (@pablogsal for that idea). I don't think we want to make 3.10/3.11 use autoconf 2.71 instead of the modified 2.69 from Christian's images at this point.

sethmlarson:
@Yhg1s That makes sense, I'll move forward with backporting that script instead.

sethmlarson:
@Yhg1s I've changed course on this since [updating the builds of 3.11](https://github.com/python/cpython/pull/114333) and prior was closed, instead now we look for the `regen-configure.sh` script and fallback to the prior implementation using Docker. Let me know what you think.

PR #89, opened 2024-02-06
Correct location of the regen-configure.sh script in 3.12+.

Yhg1s:


sethmlarson (approved):
LGTM, thanks for catching this. Having CI on this repo is looking nicer and nicer.

Issue #90, opened 2024-02-07
Strip newline off "Tool Version" in SBOM creator metadata


sethmlarson:
Current "creator" tool version can have a newline attached, we should strip the output of `git rev-parse` to prevent that in the future.

PR #92, opened 2024-02-08
Strip whitespace from commit SHA

sethmlarson:
Closes https://github.com/python/release-tools/issues/90

### DIFF
--- a/sbom.py
+++ b/sbom.py
@@ -104,7 +104,7 @@ def get_release_tools_commit_sha() -> str:
     stdout = subprocess.check_output(
         ["git", "rev-parse", "--prefix", git_prefix, "HEAD"],
         cwd=git_prefix
-    ).decode("ascii")
+    ).decode("ascii").strip()
     assert re.match(r"^[a-f0-9]{40,}$", stdout)
     return stdout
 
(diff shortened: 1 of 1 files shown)


sbom.py
```diff
@@ -104,7 +104,7 @@ def get_release_tools_commit_sha() -> str:
     stdout = subprocess.check_output(
         ["git", "rev-parse", "--prefix", git_prefix, "HEAD"],
         cwd=git_prefix
-    ).decode("ascii")
+    ).decode("ascii").strip()
     assert re.match(r"^[a-f0-9]{40,}$", stdout)
```
ezio-melotti, line 108: This `assert` is already ensuring that the `stdout` only contains the hex SHA, with no spaces.  In #90 you mentioned that sometimes there might be a newline attached, but from the report it's not clear if you saw this `assert` fail or if there was some other symptom.
Adding the `.strip()` makes no harm, so this PR is ok, but if you didn't see the assert fail, then the newline might be coming from some other place.

ezio-melotti (approved):

sbom.py
sethmlarson: I wondered about the too, @Yhg1s  did you run into this during the 3.12.2 release?

Yhg1s: Nope, I ran the release with unmodified release-tools sources, and this assert didn't trigger.

sethmlarson: Dang, I feel like I've seen a ghost. Had no idea that `$` had this behavior, now I wonder how many subtle bugs are out there.

> `$`: Matches the end of the string or just before the newline at the end of the string

PR #93, opened 2024-02-08
Add test suite and GitHub Actions CI

sethmlarson:
Closes https://github.com/python/release-tools/issues/85

PR #94, opened 2024-02-08
Add Dependabot to keep GitHub Actions up-to-date

hugovk:
Follow on from https://github.com/python/release-tools/pull/93.

Monthly, and grouped, to keep it less spammy.

Also add README with build badge.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: monthly
+    groups:
+      actions:
+        patterns:
+          - "*"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
-      - uses: actions/setup-python@65d7f2d534ac1bc67fcd62888c5f4f3d2cb2b236 # v5.0.0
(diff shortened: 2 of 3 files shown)


hugovk:
On the topic of updates, I normally use tags rather than commit hashes for actions.

And there's something a bit off about this one:

https://github.com/python/release-tools/blob/8650935702f3536b986846a345a19743d9dcecb1/.github/workflows/test.yml#L12

The comment says v5.0.0 but the [first build](https://github.com/python/release-tools/actions/runs/7836012889) says:

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/setup-python@65d7f2d534ac1bc67fcd62888c5f4f3d2cb2b236. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

However, the v5.0.0 tag https://github.com/actions/setup-python/releases/tag/v5.0.0 (https://github.com/actions/setup-python/commit/0a5c61591373683505ea898e09a3ea4f39ef2b9c, Dec 6, 2023) does say Node.js was updated to v20.

And the 65d7f2d534ac1bc67fcd62888c5f4f3d2cb2b236 commit is from Sep 7, 2023, before the upgrade.

sethmlarson:
Hmm, I used Frizbee to do the tag -> commit SHA locking. Now I'm confused if I messed up somehow or if there's an issue in that tool. Let me try reproducing that again.

sethmlarson:
I must have messed up somehow, Frizbee is resolving to `uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c # v5.0.0` correctly now. You can update that inline in this PR.

sethmlarson (approved):

PR #98, opened 2024-02-19
gh-115582 and gh-115545: Windows release build mixes up free-threaded files

zooba:


zooba:
This straightens out the builds, and also simplifies a few things now that we run so many configurations.

I've tested that the correct packages are all created for unsigned builds. Need to run a signed build to make sure nothing else goes wrong, but otherwise this should fix everything _except_ that nogil extensions won't have the right `Py_GIL_DISABLED` setting (unless they're using the Nuget packages). Because of the explicit `#undef` in pyconfig.h, they can't override it either, but that's a change to the main repo and one we should discuss in the context of all platforms (over at https://github.com/python/cpython/issues/115582).

zooba:
Also running test builds against [3.12](https://dev.azure.com/Python/cpython/_build/results?buildId=151435&view=results) and [3.10](https://dev.azure.com/Python/cpython/_build/results?buildId=151436&view=results) to make sure nothing has broken there. But that's enough for tonight.

zooba:
@ambv I don't know if you're interested in reviewing this, but it should be ready to go. It's the way I should've done the freethreaded builds in the first place - as separate configurations. Luckily, thanks to the parallelism, the overall build doesn't take much longer.

ambv:
Hm, signing binaries failed across the board here even though I did approve it.

zooba:
I'll check my Azure settings. I updated that key vault to the newer permissions model, so maybe it didn't stick properly?

zooba:
Looks like it didn't stick :) I restarted the signing jobs

zooba:
> There is a small risk doing that since you changed the "PublishPipelineArtifact@0" task into the built-in `publish` so we'll see if the checkbox is honored

Those tasks are just internal publishes. The actual checkbox impacts the entire Publish stage, so you'll see as soon as it starts running whether there's a chance it'll go public (this one won't).

zooba:
Well done finding the one combination of build options I missed 😆

ambv:
That's why they pay me the big bucks haha

ambv:
I started [another one](https://dev.azure.com/Python/cpython/_build/results?buildId=151697&view=results) to see it go to completion.

zooba:
You'll be able to grab the built files from [the Artifacts page](https://dev.azure.com/Python/cpython/_build/results?buildId=151697&view=artifacts&pathAsName=false&type=publishedArtifacts) to try them out if you want. `msi` will have the installer in it.

zooba:
@ambv Are you happy enough with the changes?

ambv:
Yeah, thanks! I went to Czechia and back over the weekend, so I didn't have access to a Windows box to test the installers. I assume they work :)

zooba:
They work at least as well as in the last alpha 😆

PR #99, opened 2024-02-22
Add SBOMs generation for Windows artifacts

sethmlarson:
Requires https://github.com/python/cpython/pull/115789 to be checked in to work.

* Does `checkout.yml` essentially pull CPython's source code into the `release-tools` directory?
* Do the Windows artifacts have pip included? I couldn't find it in `ensurepip` for the embed artifacts
* Is it possible to run Azure Pipelines off of a branch besides `master` so I can test this code before enabling it for future releases?

cc @zooba

### DIFF
--- a/windows-release/azure-pipelines.yml
+++ b/windows-release/azure-pipelines.yml
@@ -146,6 +146,12 @@ stages:
         ${{ if and(parameters.SigningCertificate, ne(parameters.SigningCertificate, 'Unsigned')) }}:
           SigningCertificate: ${{ parameters.SigningCertificate }}
 
+  - stage: SBOM
+    displayName: Create SBOMs
+    dependsOn: Build
+    jobs:
+    - template: stage-sbom.yml
+
   - stage: Layout
     displayName: Generate layouts
     dependsOn: Sign
@@ -218,7 +224,7 @@ stages:
     - ${{ if eq(parameters.DoMSI, 'true') }}:
       - stage: PublishPyDotOrg
         displayName: Publish to python.org
-        dependsOn: ['Test_MSI', 'Test']
+        dependsOn: ['SBOM', 'Test_MSI', 'Test']
         jobs:
         - template: stage-publish-pythonorg.yml
 
--- a/windows-release/stage-sbom.yml
+++ b/windows-release/stage-sbom.yml
@@ -1,0 +1,45 @@
+jobs:
+- job: SBOM_Files
+  displayName: Create SBOMs for Python binaries
+
+  pool:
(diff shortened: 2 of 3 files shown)


windows-release/stage-sbom.yml
```diff
@@ -0,0 +1,45 @@
...
+  - task: DownloadPipelineArtifact@1
+    displayName: 'Download artifact: bin_$(Name)'
+    inputs:
+      artifactName: bin_$(Name)
+      targetPath: $(Build.BinariesDirectory)\bin
+
+  - powershell: >
+      python
+      "$(Build.SourcesDirectory)\sbom.py"
+      (gci msi\*\python-*.exe | select -First 1)
```
zooba, line 37: Any reason we wouldn't/shouldn't just do all of them? (Omitting the `select -First 1` should pass them all as separate args, and then `sys.argv[1:]` in Python can get them all.)

windows-release/stage-sbom.yml
```diff
@@ -0,0 +1,45 @@
...
+    workingDirectory: $(Build.BinariesDirectory)
+    displayName: 'Create SBOMs for binaries'
+
+  - task: PublishPipelineArtifact@0
+    displayName: 'Publish artifact: sbom'
+    inputs:
+      targetPath: '$(Build.BinariesDirectory)\sbom'
+      artifactName: sbom
```
zooba, line 45:
```suggestion
  - publish: '$(Build.BinariesDirectory)\sbom'
    artifact: sbom
    displayName: 'Publish artifact: sbom'
```

This is the preferred format for simple cases now (should auto-update when they need to make changes to the publish task).

windows-release/azure-pipelines.yml
```diff
@@ -218,7 +224,7 @@ stages:
     - ${{ if eq(parameters.DoMSI, 'true') }}:
       - stage: PublishPyDotOrg
         displayName: Publish to python.org
-        dependsOn: ['Test_MSI', 'Test']
+        dependsOn: ['SBOM', 'Test_MSI', 'Test']
         jobs:
         - template: stage-publish-pythonorg.yml
```
zooba, line 229: I assume we're going to get chances to this template as well to SSH the files up to the server?

sethmlarson: Yeah that's the plan, I might do that in a follow-up PR though.

sethmlarson:
Thanks for the review @zooba, I've closed this PR and [created a new one](https://github.com/python/release-tools/pull/100) on this repo for easier testing and addressed your comments in the new PR.

PR #100, opened 2024-02-23
 Add SBOMs generation for Windows artifacts

sethmlarson:
[Moved the branch to this repo](https://github.com/python/release-tools/pull/99) to allow testing before merging in Azure Pipelines. Requires https://github.com/python/cpython/pull/115789 to be checked in to work.


windows-release/stage-sbom.yml
```diff
@@ -0,0 +1,53 @@
...
+    displayName: 'Download artifact: bin_$(Name)'
+    inputs:
+      artifactName: bin_$(Name)
+      targetPath: $(Build.BinariesDirectory)\bin
+
+  - powershell: >
+      python
+      "$(Build.SourcesDirectory)\sbom.py"
+      "--cpython-source-dir=$(Build.SourcesDirectory)"
+      (gci msi\*\python-*.exe)
```
zooba, line 38: The MSIs are not available until the very end of the build, and only if the MSI stage ran (which is optional).

Perhaps a better place to put this is in the `layout` stages? Those should be platform-specific and package-specific already, so you can pass options to the script directly.

sethmlarson: I took a stab at this in https://github.com/python/release-tools/pull/100/commits/1d23d2da532a389877b86a6c88ad5f29838bcb26, not certain if this is what you were imagining. Also wasn't sure if `$(Python)` would be available in these contexts and if I'm using `gci` correctly. My Windows knowledge is fairly limited :grimacing:

windows-release/msi-steps.yml
```diff
@@ -114,3 +114,26 @@ steps:
   - publish: '$(Build.ArtifactStagingDirectory)\msi'
     artifact: msi
     displayName: 'Publish MSI'
+
+  - powershell: >
+      $(Python)
```
zooba, line 119:
```suggestion
      & $(Python)
```

I'm 99% sure you'll need this. The Python variable here has been exported by an earlier step that gets it from one of our scripts that assumes it's running in `cmd.exe`, so it's possible that the variable contains a quoted path. In Powershell, this looks like a string literal, so it won't treat it as a quoted command unless you add the `&` (which, IIUC, just means "treat this entire input as a command").

windows-release/msi-steps.yml
```diff
@@ -114,3 +114,26 @@ steps:
...
+    displayName: 'Create SBOMs for binaries'
+
+  - task: CopyFiles@2
+    displayName: 'Layout Artifact: sbom'
+    inputs:
+      sourceFolder: $(Build.ArtifactStagingDirectory)\msi
+      targetFolder: $(Build.ArtifactStagingDirectory)\sbom
+      flatten: true
+      contents: |
+        **\*.spdx.json
```
zooba, line 133: Would be nice to put the above two steps before the previous `publish` step. It's uncommon, but when a task fails for no apparent reason, we can usually retry the job. However, if anything has been published, the publish step will fail and we need to restart the entire build. So ideally, we want to do all the work first and then publish.

I don't have any concerns about including the SBOMs in the msi artifact.

windows-release/stage-layout-embed.yml
```diff
@@ -55,3 +55,26 @@ jobs:
     inputs:
       PathtoPublish: '$(Build.ArtifactStagingDirectory)\embed'
       ArtifactName: embed
+
+  - powershell: >
+      $(Python)
```
zooba, line 60: This one shouldn't need the `&` because it's been set by `layout-command.yml` (but I think we're just hoping that we never have a space in the path...)

One possibility is this (which is how `layout-command.yml` uses it), but it probably makes no difference:

```suggestion
      & "$(Python)"
```

zooba (commented):
I'm confident enough we need the `&` in there that I'll wait for that change before running another build.

You're using `gci` fine, btw (it's short for `Get-ChildItem` which is also aliased as `dir` and `ls`, because who needs only one way to do things...)

windows-release/stage-layout-embed.yml
sethmlarson: Done in https://github.com/python/release-tools/pull/100/commits/c01765dacd7876de68e712362e2f8015aafae02b

windows-release/msi-steps.yml
sethmlarson: Done in https://github.com/python/release-tools/pull/100/commits/c01765dacd7876de68e712362e2f8015aafae02b, also moved the positioning in the other job so publishing is last.

windows-release/msi-steps.yml
sethmlarson: Done in https://github.com/python/release-tools/pull/100/commits/c01765dacd7876de68e712362e2f8015aafae02b, thanks for the Powershell knowledge!

zooba:
New test build running at https://dev.azure.com/Python/cpython/_build/results?buildId=152746&view=results

windows-release/stage-layout-embed.yml
```diff
@@ -46,6 +46,23 @@ jobs:
       --preset-embed
     displayName: 'Generate embeddable layout'
 
+  - powershell: >
+      & "$(Python)"
+      "$(Build.SourcesDirectory)\sbom.py"
```
zooba, line 51: Okay, so I guess we don't have this file without doing a `checkout: self` step first, which is going to mess with every other step and template (because we currently don't checkout multiple repositories anywhere, so they aren't "aware" of it).

It might be easiest to add a custom `git clone` command to clone our own repo into its own directory. I _think_:

```
git clone $(Build.Repository.Uri) -b $(Build.SourceBranch) --single-branch --no-checkout "$(Pipeline.Workspace)\release-tools"
git -C "$(Pipeline.Workspace)\release-tools" checkout $(Build.SourceVersion)
```

And then:
```suggestion
      "$(Pipeline.Workspace)\release-tools\sbom.py"
```

sethmlarson: Got it! Done in https://github.com/python/release-tools/pull/100/commits/04ee7a65197cdb2b6a884c66283f18cea7ea3851

sethmlarson:
Thanks for the run, the joys of developing CI workflow definitions continues :)

windows-release/stage-layout-embed.yml
```diff
@@ -46,6 +46,28 @@ jobs:
       --preset-embed
     displayName: 'Generate embeddable layout'
 
+  - powershell: >
```
zooba, line 49:
```suggestion
  - powershell: |
```

Separate commands, so we need them to keep the line breaks.

windows-release/msi-steps.yml
```diff
@@ -111,6 +111,29 @@ steps:
           *.cab
           *.exe
 
+  - powershell: >
+      & $(Python)
+      "$(Build.SourcesDirectory)\sbom.py"
```
zooba, line 116: Looks like we need the extra clone in this build too.

zooba:
Looking at the most recent failure, it seems you probably want `$(Build.SourceBranchName)` rather than `$(Build.SourceBranch)`. (I really wish Git had a better way to just clone a single known commit, but apparently not...)

sethmlarson:
The SBOM artifacts are getting uploaded into the `sbom` artifact name as expected: https://dev.azure.com/Python/cpython/_build/results?buildId=152774&view=artifacts&pathAsName=false&type=publishedArtifacts

I downloaded them all and gave them a look, they contain the components I expect and SBOM tooling accepts them. This PR can be ready to go as-is or we could also add the "upload" step in this PR too.

zooba (approved):
I _think_ we should just be able to add a `*.spdx.json` wildcard to `$exe` in `uploadrelease.ps1` to pick up the SBOMs, right? Or maybe add a new parameter like `$embed` to specify the directory, to save copying them into the main directory.

TBH, I don't _love_ the whole upload script, I just haven't looked at it in years (and it was written for a local build, which doesn't look like the automated one). Take a look and see whether you have a desperate desire to rewrite it, in which case we can do a new PR, or if you can easily hack it in then we can do it now.

PR #101, opened 2024-02-27
Run Windows installer tests in their own container

zooba:
This makes each install test run in its own container, isolated from the build machine, which allows us to test versions of Python that are already installed on the build machine.

Unfortunately, it seems that Windows Installer does not work properly inside the container, so we can't actually use it.

Posting it here for posterity, or in case someone comes up with a way to make it work.

### DIFF
--- a/windows-release/test-msi.ps1
+++ b/windows-release/test-msi.ps1
@@ -27,16 +27,24 @@ dir -r "${env:APPDATA}\Microsoft\Windows\Start Menu\Programs\Python*"
 
 Write-Host "##[section]Capture registry"
 Write-Host 'Capture per-machine 32-bit registry'
-Write-Host "##[command]dir -r ""HKLM:\Software\WOW6432Node\Python"""
-dir -r "HKLM:\Software\WOW6432Node\Python"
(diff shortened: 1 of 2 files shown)


windows-release/test-msi.ps1
```diff
@@ -0,0 +1,100 @@
...
+dir -r "${env:APPDATA}\Microsoft\Windows\Start Menu\Programs\Python*"
+
+Write-Host "##[section]Capture registry"
+Write-Host 'Capture per-machine 32-bit registry'
+# PS 5.0 and later can't do this, because they normalise registry paths incorrectly
+# So we'll use the good old "reg" tool
+#Write-Host "##[command]dir -r ""HKLM:\Software\WOW6432Node\Python"""
+#dir -r "HKLM:\Software\WOW6432Node\Python"
+Write-Host "##[command]reg HKLM\Software\Python /s /reg:32"
+reg HKLM\Software\Python /s /reg:32
```
zooba, line 35: Also, the containers don't seem to have `reg.exe`, so these steps probably need to be replaced with our own custom code for listing the registry.

Or maybe we can downgrade Powershell to a version that isn't broken by https://github.com/PowerShell/PowerShell/issues/5536

PR #104, opened 2024-04-02
Correct doc detection in GH release run

pablogsal:

PR #105, opened 2024-04-02
Triple lint YAML

hugovk:
Follow on from https://github.com/python/release-tools/pull/104.

Before that fix, the "Upload the docs artifacts" step was being skipped:

https://github.com/python/release-tools/actions/runs/8519755747/job/23334443184

Linting with https://github.com/rhysd/actionlint:

```console
❯ actionlint
.github/workflows/source-and-docs-release.yml:71:9: shellcheck reported issue in this script: SC2046:warning:7:22: Quote this to prevent word splitting [shellcheck]
   |
71 |         run: |
   |         ^~~~
.github/workflows/source-and-docs-release.yml:96:13: if: condition "${{ hashFiles('cpython/${{ inputs.cpython_release }}/docs') != '' }}" is always evaluated to true because extra characters are around ${{ }} [if-cond]
   |
96 |         if: ${{ hashFiles('cpython/${{ inputs.cpython_release }}/docs') != '' }}
   |             ^~~
```

That second one was the problem: it doesn't like the `${{ inputs.cpython_release }` inside a string: 'cpython/${{ inputs.cpython_release }}/docs' (try in the [playground](https://rhysd.github.io/actionlint)).

See:

* https://github.com/orgs/community/discussions/25761#discussioncomment-6306201

Let's also fix that shellcheck warning, and add _three_ YAML linters to the CI to help prevent this :)  Plus some other handy linters.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,21 @@
+name: Lint
+
+on: [push, pull_request, workflow_dispatch]
+
+env:
+  FORCE_COLOR: 1
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.x"
+          cache: pip
+      - uses: pre-commit/action@v3.0.1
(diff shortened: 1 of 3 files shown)


pablogsal (approved):
Linters for the linters god!

sethmlarson (approved):
LGTM, thanks Hugo!

PR #109, opened 2024-04-09
Add more linting

hugovk:
Any thoughts about adding Black in a future PR?


JelleZijlstra:
Might also be worth adding a type checker, which could have caught #107.

hugovk:
Good idea, I've opened https://github.com/python/release-tools/issues/114 for that.

---

Conflict resolved.

add-to-pydotorg.py
```diff
@@ -206,36 +205,36 @@ def list_files(release):
...
         else:
-            print('    File %s/%s not recognized' % (reldir, rfile))
+            print(f'    File {reldir}/{rfile} not recognized')
             continue
 
 def query_object(objtype, **params):
     """Find an API object by query parameters."""
     uri = base_url + 'downloads/%s/' % objtype
-    uri += '?' + '&'.join('%s=%s' % v for v in params.items())
+    uri += '?' + '&'.join('{}={}'.format(*v) for v in params.items())
```
ezio-melotti, line 234:
```suggestion
    uri += '?' + '&'.join(f'{k}={v}' for k, v in params.items())
```

release.py
```diff
@@ -168,12 +167,12 @@ def tweak_patchlevel(tag, done=False):
...
-        b   = 'PY_RELEASE_LEVEL_BETA',
-        rc  = 'PY_RELEASE_LEVEL_GAMMA',
-        f   = 'PY_RELEASE_LEVEL_FINAL',
-        )[tag.level]
+    level_def = {
+        'a': 'PY_RELEASE_LEVEL_ALPHA',
+        'b': 'PY_RELEASE_LEVEL_BETA',
+        'rc': 'PY_RELEASE_LEVEL_GAMMA',
+        'f': 'PY_RELEASE_LEVEL_FINAL',
+        }[tag.level]
```
ezio-melotti, line 175:
```suggestion
    level_def = {
        'a': 'PY_RELEASE_LEVEL_ALPHA',
        'b': 'PY_RELEASE_LEVEL_BETA',
        'rc': 'PY_RELEASE_LEVEL_GAMMA',
        'f': 'PY_RELEASE_LEVEL_FINAL',
    }[tag.level]
```
Shouldn't this line be dedented?  Maybe the linter gets confused by the `[tag.level]`?

release.py
```diff
@@ -370,7 +369,7 @@ def build_docs():
     with tempfile.TemporaryDirectory() as venv:
         run_cmd(['python3', '-m', 'venv', venv])
         pip = os.path.join(venv, 'bin', 'pip')
-        run_cmd([pip, 'install', '-r' 'Doc/requirements.txt'])
+        run_cmd([pip, 'install', '-rDoc/requirements.txt'])
```
ezio-melotti, line 372:
```suggestion
        run_cmd([pip, 'install', '-r', 'Doc/requirements.txt'])
```

release.py
```diff
@@ -421,16 +420,16 @@ def __init__(self, tag_name):
         self.level = data[3]
         self.serial = int(data[4])
         # This has the effect of normalizing the version.
-        self.text = "{}.{}.{}".format(self.major, self.minor, self.patch)
+        self.text = f"{self.major}.{self.minor}.{self.patch}"
```
ezio-melotti, line 423:
```suggestion
        self.text = self.normalized()
```
This might be out of the scope of this PR, since it's not linting-related but about removing duplication, so feel free to ignore or fix in a separate PR.  The comment above might also need to be updated.

.ruff.toml
```diff
@@ -0,0 +1,23 @@
...
+  "UP",     # pyupgrade
+  "W",      # pycodestyle warnings
+  "YTT",    # flake8-2020
+]
+ignore = [
+  "E203",   # Whitespace before ':'
+  "E221",   # Multiple spaces before operator
+  "E226",   # Missing whitespace around arithmetic operator
+  "E241",   # Multiple spaces after ','
+  "E501",   # Line too long
```
ezio-melotti, line 22: Why are these ignored?

release.py
hugovk: I would personally dedent it but the linter isn't bothered.

I avoided making any formatting changes here, we should just run Black on everything.

release.py
hugovk: Yep, I'll add it to https://github.com/python/release-tools/pull/110 because it's also adding tests here 👍

.ruff.toml
hugovk: E2xxx because they're some common ones that are incompatible with Black. I copied these over from another project along with the `select` list. We could remove them here until adding Black if you like? They don't flag anything right now.

E501 because there's lots of long lines. We can remove this when we add Black formatting.

sethmlarson (approved):

hugovk:
(Conflict resolved)

sethmlarson (approved):

PR #117, opened 2024-04-23
Upload Windows SBOMs to python.org

sethmlarson:
I wasn't quite sure how to test these changes since we don't actually want to upload to python.org.

### DIFF
--- a/windows-release/stage-publish-pythonorg.yml
+++ b/windows-release/stage-publish-pythonorg.yml
@@ -57,6 +57,17 @@ jobs:
         buildVersionToDownload: specific
         buildId: ${{ parameters.BuildToPublish }}
 
+    - task: DownloadBuildArtifacts@0
+      displayName: 'Download artifact from ${{ parameters.BuildToPublish }}: sbom'
+      inputs:
+        artifactName: sbom
+        downloadPath: $(Build.BinariesDirectory)\sbom
+        buildType: specific
+        project: $(System.TeamProject)
+        pipeline: $(Build.DefinitionName)
+        buildVersionToDownload: specific
+        buildId: ${{ parameters.BuildToPublish }}
+
   - ${{ else }}:
     - task: DownloadPipelineArtifact@1
       displayName: 'Download artifact: Doc'
@@ -77,6 +88,12 @@ jobs:
         artifactName: embed
         downloadPath: $(Build.BinariesDirectory)
 
+    # Note that sbom is a 'build' artifact, not a 'pipeline' artifact
+    - task: DownloadBuildArtifacts@0
+      displayName: 'Download artifact: sbom'
+      inputs:
+        artifactName: sbom
+        downloadPath: $(Build.BinariesDirectory)\sbom
 
   # Note that ARM64 MSIs are skipped at build when this option is specified
   - powershell: 'gci *embed-arm*.zip | %{ Write-Host "Not publishing: $($_.Name)"; gi $_ } | del'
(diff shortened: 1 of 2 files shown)


windows-release/stage-publish-pythonorg.yml
```diff
@@ -57,6 +57,17 @@ jobs:
         buildVersionToDownload: specific
         buildId: ${{ parameters.BuildToPublish }}
 
+    - task: DownloadBuildArtifacts@0
+      displayName: 'Download artifact from ${{ parameters.BuildToPublish }}: sbom'
+      inputs:
+        artifactName: sbom
+        downloadPath: $(Build.BinariesDirectory)\sbom
```
sethmlarson, line 64: Does `DownloadBuildArtifact` create the directory at `downloadPath`?

windows-release/uploadrelease.ps1
```diff
@@ -87,19 +88,28 @@ $dirs = gci "$build" -Directory
...
     $msi = gci *.msi, *.msi.asc, *.msu, *.msu.asc
+    $spdx_json = gci *.spdx.json
     popd
 
     if ($exe) {
         & $pscp -batch -hostkey $hostkey -noagent -i $keyfile $exe.FullName "$user@${server}:$d"
         if (-not $?) { throw "Failed to upload $exe" }
     }
 
+    if ($spdx_json) {
```
sethmlarson, line 108: This probably isn't right, the SBOM directory contains more than one file, but I believe other artifacts are like that too so I duplicated what was done for exe?

windows-release/uploadrelease.ps1
```diff
@@ -87,19 +88,28 @@ $dirs = gci "$build" -Directory
 if ($embed) {
     $dirs = ($dirs, (gi $embed)) | %{ $_ }
 }
+if ($sbom) {
+    $dirs = ($dirs, $sbom)
```
zooba, line 92:
```suggestion
    $dirs = ($dirs, $sbom) | %{ $_ }
```

Theoretically, the last bit shouldn't be needed (it serves to flatten the two arrays), but I'm sure I put it there on line 88/89 for a reason.

windows-release/uploadrelease.ps1
```diff
@@ -87,19 +88,28 @@ $dirs = gci "$build" -Directory
...
     popd
 
     if ($exe) {
         & $pscp -batch -hostkey $hostkey -noagent -i $keyfile $exe.FullName "$user@${server}:$d"
         if (-not $?) { throw "Failed to upload $exe" }
     }
 
+    if ($spdx_json) {
+        & $pscp -batch -hostkey $hostkey -noagent -i $keyfile $spdx_json.FullName "$user@${server}:$d"
+        if (-not $?) { throw "Failed to upload $spdx_json" }
```
zooba, line 110:
```suggestion
        if (-not $?) { Write-Host "##[warning]Failed to upload $spdx_json" }
```

At least until we're confident this is going to work, I'd rather not break the next release, especially after files have already been uploaded. All a break here does is make it highly likely that we'll end up with a second build and mismatched binaries between sources (e.g. Nuget will have already published by this point).

zooba (commented):
Two changes and I think this is good to merge. We can't really test until a release comes around, but I guess 3.13b1 isn't far off now.

sethmlarson:
Thanks for the review @zooba! I applied your suggestions :)

zooba:
Alright, we're in! Will find out what happens when we release next...

PR #119, opened 2024-04-25
Upgrade dependencies with vulnerabilities, require binaries

sethmlarson:
* Makes all dependencies except grpclib and grapheme binary-only to avoid compromise via build backend.
* Upgrades cryptography (and pyOpenSSL) and pydantic


hugovk (approved):
> except grpclib and grapheme

Because they only have sdists, not wheels.

* grpclib should have a wheel for the next release: https://github.com/vmagamedov/grpclib/pull/188

* grapheme previously had wheels for 0.2.1-0.5.0 but not the latest 0.6.0: https://pypi.org/project/grapheme/#history. I opened an issue to ask for the 0.6.0 wheel.

sethmlarson:
Thanks @hugovk!

PR #120, opened 2024-04-26
Separate source and documentation builds

### LINKED ISSUE #74: Separate builds for source and docs due to dependency differences
The builds for the documentation need to pull in many more dependencies, from both PyPI and apt. Once the source and docs builds have been moved to GitHub Actions the two should be separated so as not to allow a compromise of a docs dependency to affect the more important source artifact builds.

Requires: https://github.com/python/release-tools/issues/66

From naively building a Docker container for ubuntu:22.04 and running the install scripts as they are in GitHub Actions we see (using Syft):

* 843 dependencies for docs and source: [source-and-docs-deps.txt](https://github.com/python/release-tools/files/15131388/source-and-docs-deps.txt)
* 171 dependencies for source [source-deps.txt](https://github.com/python/release-tools/files/15131391/source-deps.txt)

sethmlarson:
This PR separates the builds for the source artifacts and documentation into separate jobs. The primary motivation is that the docs builds require pulling many more (mostly unpinned) dependencies from PyPI and apt in order to generate the documentation whereas the source artifacts which have higher integrity requirements require far fewer, pinned, dependencies.

Originally when I looked at this problem it seems somewhat difficult because the docs builds and source builds shared a build script, but after [running all of the current release streams](https://github.com/python/release-tools/issues/74#issuecomment-2079425442) I'm not sure why the script has the docs build step as a part of the source build step. All builds seemed to succeed without building the docs and source together.

If it's possible to remove the docs build from `release.py --export` then I can update the commit and merge this PR.

Closes #74

### DIFF
--- a/.github/workflows/source-and-docs-release.yml
+++ b/.github/workflows/source-and-docs-release.yml
@@ -19,7 +19,7 @@ on:
 name: "Build Python source and docs artifacts"
 
 jobs:
-  build-source-and-docs:
+  verify-input:
     runs-on: ubuntu-22.04
     steps:
       - name: "Workflow run information"
@@ -27,9 +27,6 @@ jobs:
           echo "git_remote: ${{ inputs.git_remote }}"
           echo "git_commit: ${{ inputs.git_commit }}"
           echo "cpython_release: ${{ inputs.cpython_release }}"
-
-      - name: "Checkout python/release-tools"
-        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
 
       - name: "Checkout ${{ inputs.git_remote }}/cpython"
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
@@ -45,6 +42,21 @@ jobs:
             exit 1
           fi
 
+  build-source:
+    runs-on: ubuntu-22.04
+    needs:
+      - verify-input
+    steps:
+      - name: "Checkout python/release-tools"
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: "Checkout ${{ inputs.git_remote }}/cpython"
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        with:
+          repository: "${{ inputs.git_remote }}/cpython"
+          ref: "v${{ inputs.cpython_release }}"
+          path: "cpython"
+
       - name: "Setup Python"
         uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
         with:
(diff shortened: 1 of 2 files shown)


release.py
```diff
@@ -315,8 +315,8 @@ def export(tag, silent=False):
             # build docs *before* we do "blurb export"
             # because docs now depend on Misc/NEWS.d
             # and we remove Misc/NEWS.d as part of cleanup for export
-            if tag.is_final or tag.level == 'rc':
-                docdist = build_docs()
+            # if tag.is_final or tag.level == 'rc':
+            #     docdist = build_docs()
```
hugovk, line 319: Remember to uncomment before merge:
```suggestion
            if tag.is_final or tag.level == 'rc':
                docdist = build_docs()
```

release.py
```diff
@@ -337,8 +337,8 @@ def export(tag, silent=False):
             for name in ('.azure-pipelines', '.git', '.github', '.hg'):
                 shutil.rmtree(name, ignore_errors=True)
 
-        if tag.is_final or tag.level == 'rc':
-            shutil.copytree(docdist, 'docs')
+        # if tag.is_final or tag.level == 'rc':
+        #     shutil.copytree(docdist, 'docs')
```
hugovk, line 341:
```suggestion
        if tag.is_final or tag.level == 'rc':
            shutil.copytree(docdist, 'docs')
```

.github/workflows/source-and-docs-release.yml
```diff
@@ -80,20 +79,53 @@ jobs:
...
+      - name: "Checkout ${{ inputs.git_remote }}/cpython"
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        with:
+          repository: "${{ inputs.git_remote }}/cpython"
+          ref: "v${{ inputs.cpython_release }}"
+
+      - name: "Setup Python"
+        uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
+        with:
+          python-version: 3.11
```
hugovk, line 100: We could bump this and the other to 3.12 at some point?

hugovk (approved):
I don't see any problems with the before and after artifacts.

The source zip contents are binary identical.

The docs have different timestamps in the HTML and different theme versions (expected). I quickly compared visually a couple of PDFs and they looked fine.

.github/workflows/source-and-docs-release.yml
sethmlarson: Agreed, I can submit a PR for that too.

sethmlarson:
@hugovk Definitely don't want to keep the steps commented out, but how would you like to proceed on skipping the docs build? Perhaps adding a new option for `--skip-docs`? to `release.py`?

hugovk:
Yep, `--skip-docs` sounds good for me.

sethmlarson:
@hugovk Done in https://github.com/python/release-tools/pull/120/commits/cc7521788c2ad31f9935dff59560a1be6d99870e, I'll run the action again to verify it works.

release.py
```diff
@@ -514,6 +519,8 @@ def main(argv):
     chdir_to_repo_root()
     parser = get_arg_parser()
     options, args = parser.parse_args(argv)
+    if options.skip_docs and not options.export:
+        error("--skip-docs option has no effect without --export")
```
hugovk, line 523: I was going to suggest using [`add_mutually_exclusive_group`](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group) instead, but then saw we're using the deprecated optparse rather than argparse!

Something for another time :)

hugovk (approved):
Thanks!

PR #121, opened 2024-04-30
Rearrange pip format requirements

sethmlarson:
Noticed that pip-compile doesn't preserve the behavior of `--no-binary`/`--only-binary` so our `requirements.txt` file actually couldn't be installed with `pip install`.

I [opened an issue](https://github.com/jazzband/pip-tools/issues/2081) and [fix](https://github.com/jazzband/pip-tools/pull/2082) with pip-tools that fixes the issue and manually amended our `requirements.txt` file. Also moved the format requirements to `requirements.in` so we can document the situation for each dependency that requires an sdist.


webknjaz:
@sethmlarson additionally, pip-tools supports a config file for almost a year so you could save some common args there.

hugovk (approved):

PR #128, opened 2024-05-08
Update release tools to regenerate ``configure`` and some files for Windows build

### LINKED ISSUE #45: Post-release-branch update of main fails to update a few things
The post-release-branch update of main, where it bumps the Python version number, failed to update a few things, which broke CI (https://github.com/python/cpython/actions/runs/5049581081). At the very least:
 - configure wasn't regenerated after configure.ac was updated.
 - PC/pyconfig.h wasn't updated with the right .lib files (https://github.com/python/cpython/pull/104755)
 
There also seem to be a few test failures from the version bump, which may or may not be the release script's fault, but certainly should have been caught before pushing to main. The release script should probably update main and run its tests while waiting for the binary artifacts...

### LINKED ISSUE #126: run_release should regenerate configure after editing configure.ac.
As part of the release process for beta 1, run_release edits (or makes the user edit, rather) in `main` various files to replace the just-released version with the next major version. One of the files that needs to be changed is `configure.ac`, but run_release doesn't then regenerate the configure script, causing all builds to fail when pushed to github.

Eclips4:
Fixes #126 and Fixes #45


run_release.py
```diff
@@ -1138,10 +1138,10 @@ def _api_key(api_key):
         Task(run_blurb_release, "Run blurb release"),
         Task(check_cpython_repo_is_clean, "Checking git repository is clean"),
         Task(prepare_pydoc_topics, "Preparing pydoc topics"),
+        Task(bump_version, "Bump version"),
```
Eclips4, line 1141: Previously, regeneration of ``configure`` was running before ``bump_version`, so it actually "worked", but not as expected :)

ambv: Yeah, this is fine, but note that this was paired with the "is repository clean" check in the line right after. I'd move that, too.

Eclips4: Sure. Check the last commit, please.

hugovk:
> Fixes #126 and #45

GitHub tip: you need to repeat the magic "Fixes" keyword before each issue number for it to autoclose both (you can mouseover the keyword to see what it will do).

ambv:
Thanks, Kirill!

Issue #114, opened 2024-04-10
Add a type checker


hugovk:
Might also be worth adding a type checker, which could have caught #107.

_Originally posted by @JelleZijlstra in https://github.com/python/release-tools/issues/109#issuecomment-2045309417_

PR #131, opened 2024-05-26
Add type hints to release.py and run mypy on CI

hugovk:
Make a start on https://github.com/python/release-tools/issues/114.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,6 +8,7 @@ follow_imports = "silent"
 enable_error_code = "ignore-without-code"
 extra_checks = true
 warn_unreachable = true
+
 exclude = [
   "^add-to-pydotorg.py$",
   "^buildbotapi.py$",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,6 @@
 python_version = "3.12"
 pretty = true
 strict = true
-follow_imports = "silent"
 
 # Extra checks that aren't included in --strict
 enable_error_code = "ignore-without-code"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,20 @@
+[tool.mypy]
+python_version = "3.12"
+pretty = true
+disallow_any_generics = true
+enable_error_code = "ignore-without-code"
+extra_checks = true
+follow_imports = "silent"
+strict = true
+warn_redundant_casts = true
+warn_unreachable = true
+warn_unused_ignores = true
+exclude = [
+  "^add-to-pydotorg.py$",
+  "^buildbotapi.py$",
+  "^run_release.py$",
+  "^sbom.py$",
+  "^tests/test_release_tag.py$",
+  "^tests/test_sbom.py$",
+  "^windows-release/purge.py$",
+]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ pretty = true
 strict = true
 
 # Extra checks that aren't included in --strict
-enable_error_code = "ignore-without-code"
+enable_error_code = "ignore-without-code,redundant-expr,truthy-iterable"
 extra_checks = true
 warn_unreachable = true
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,14 +1,13 @@
 [tool.mypy]
 python_version = "3.12"
 pretty = true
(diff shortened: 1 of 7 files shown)


release.py
```diff
@@ -71,7 +160,10 @@ def chdir_to_repo_root():
 
         os.chdir(path)
 
-        def test_first_line(filename, test):
+        def test_first_line(
+            filename: str,
+            test: Callable[[str], re.Match[str] | None] | Callable[[object], bool],
```
JelleZijlstra, line 165:
```suggestion
            test: Callable[[str], object],
```
Avoid overly specific type.

release.py
```diff
@@ -201,6 +295,7 @@ def tweak_patchlevel(tag, done=False):
 
 /* Version as a string */
 #define PY_VERSION      \t\"{tag.text}{plus}"'''.strip()
+    assert tag.level is str
```
JelleZijlstra, line 298: Do you mean `isinstance`?

tox.ini
```diff
@@ -0,0 +1,38 @@
...
+    pre-commit
+pass_env =
+    PRE_COMMIT_COLOR
+commands =
+    pre-commit run --all-files --show-diff-on-failure
+
+[testenv:mypy]
+skip_install = true
+deps =
+    mypy==1.10
```
JelleZijlstra, line 34: Will dependabot update this?

pyproject.toml
```diff
@@ -0,0 +1,20 @@
...
+python_version = "3.12"
+pretty = true
+disallow_any_generics = true
+enable_error_code = "ignore-without-code"
+extra_checks = true
+follow_imports = "silent"
+strict = true
+warn_redundant_casts = true
+warn_unreachable = true
+warn_unused_ignores = true
```
AlexWaygood, line 11: A bunch of these are already enabled anyway due to `strict = true`:

```suggestion
python_version = "3.12"
pretty = true
strict = true
follow_imports = "silent"

# Extra checks that aren't included in --strict
enable_error_code = "ignore-without-code"
extra_checks = true
warn_unreachable = true
```

What's the reason for using `follow_imports = "silent"` rather than using the default setting here?

hugovk: I probably just copied and pasted from some other project :)

tox.ini
hugovk: It won't, I'll move it to a requirements file.

release.py
hugovk: Yep, will update.

pyproject.toml
AlexWaygood: Cool, I think most of the `follow_imports` options do terrible/unexpected things in some situations, so I'd definitely rather stick with the default here if possible!! If there are specific imports that are causing issues and we need a non-default setting, I'd rather we add config for them individually, e.g. like we do in https://github.com/PyCQA/flake8-pyi/blob/f99a1b1dfe713706b3d413a69dd231b9cd2de676/pyproject.toml#L107-L109

pyproject.toml
```diff
@@ -0,0 +1,19 @@
...
+python_version = "3.12"
+pretty = true
+strict = true
+follow_imports = "silent"
+
+# Extra checks that aren't included in --strict
+enable_error_code = "ignore-without-code"
+extra_checks = true
+warn_unreachable = true
+exclude = [
```
AlexWaygood, line 11: oops, I guess a newline here would be helpful to clarify that this isn't an "extra check" (the comment only applies to the above block)

```suggestion

exclude = [
```

pyproject.toml
```diff
@@ -0,0 +1,19 @@
+[tool.mypy]
+python_version = "3.12"
+pretty = true
+strict = true
+
+# Extra checks that aren't included in --strict
+enable_error_code = "ignore-without-code"
```
AlexWaygood, line 7: These checks can also often be useful (but only enable them if they actually pass, which I haven't checked!)

```suggestion
enable_error_code = "ignore-without-code,redundant-expr,possibly-undefined,truthy-iterable"
```

AlexWaygood (approved):

pyproject.toml
AlexWaygood: Either fix the complaint or remove it from the list, I leave it to you ;)

PR #133, opened 2024-06-05
Restore the order of pip options

encukou:
This makes `--no-binary grapheme` effective.

This partially reverts 28583c1168939394ba72105a5801fd12f23df64e,
and restores c9f4eca9fb7b6c2d0eed9bf0c929fceb6e4b1d42.

I've only tested locally with `pip install -r requirements.txt`.

PR #134, opened 2024-06-05
Test running "Build Python source and docs artifacts" workflow

hugovk:
PR https://github.com/python/release-tools/pull/133 fixed a problem caused by PR https://github.com/python/release-tools/pull/130: Dependabot moved the `--only-binary :all:` option. I merged it because the CI passed. But the CI didn't actually run this workflow.

This PR adds triggers to run this for push/PR, in addition to the existing manual trigger (`workflow_dispatch`). In these cases, we don't have inputs (`github.event.inputs`), so set some default values instead.

We can skip running it when some test/lint files are changed, like `.github/dependabot.yml` and `tests/` files.

Here's a sample PR trigger:

* https://github.com/python/release-tools/actions/runs/9385278500/job/25843041283?pr=134

Here's a sample manual trigger:

* https://github.com/hugovk/release-tools/actions/runs/9385143957

### DIFF
--- a/.github/workflows/source-and-docs-release.yml
+++ b/.github/workflows/source-and-docs-release.yml
@@ -64,8 +64,8 @@ jobs:
 
       - name: "Verify CPython commit matches tag"
         run: |
-          if [[ "${{ env.GIT_COMMIT }}" != "$(cd cpython && git rev-parse HEAD)" ]]; then
-            echo "expected git commit ('${{ env.GIT_COMMIT }}') didn't match tagged commit ('$(git rev-parse HEAD)')"
+          if [[ "$GIT_COMMIT" != "$(cd cpython && git rev-parse HEAD)" ]]; then
+            echo "expected git commit ('$GIT_COMMIT') didn't match tagged commit ('$(git rev-parse HEAD)')"
             exit 1
           fi
 
@@ -97,7 +97,7 @@ jobs:
       - name: "Build Python release artifacts"
         run: |
           cd cpython
-          python ../release.py --export ${{ env.CPYTHON_RELEASE }} --skip-docs
+          python ../release.py --export $CPYTHON_RELEASE --skip-docs
 
       - name: "Upload the source artifacts"
         uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
@@ -161,10 +161,10 @@ jobs:
       - name: "Test Python source tarballs"
         run: |
           mkdir -p ./tmp/installation/
-          cp Python-${{ env.CPYTHON_RELEASE }}.tgz ./tmp/
+          cp Python-$CPYTHON_RELEASE.tgz ./tmp/
           cd tmp/
-          tar xvf Python-${{ env.CPYTHON_RELEASE }}.tgz
-          cd Python-${{ env.CPYTHON_RELEASE }}
+          tar xvf Python-$CPYTHON_RELEASE.tgz
+          cd Python-$CPYTHON_RELEASE
 
           ./configure "--prefix=$(realpath '../installation/')"
           make -j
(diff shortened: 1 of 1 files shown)


.github/workflows/source-and-docs-release.yml
```diff
@@ -18,27 +36,36 @@ on:
 
 name: "Build Python source and docs artifacts"
 
+# Set from inputs for workflow_dispatch, or set defaults to test push/PR events
+env:
+  GIT_REMOTE: ${{ github.event.inputs.git_remote || 'python' }}
+  GIT_COMMIT: ${{ github.event.inputs.git_commit || '2268289a47c6e3c9a220b53697f9480ec390466f' }}
+  CPYTHON_RELEASE: ${{ github.event.inputs.cpython_release || '3.13.0b1' }}
```
hugovk, line 43: Perhaps default to a non-alpha/non-beta, so a docs build is also run?

.github/workflows/source-and-docs-release.yml
```diff
@@ -70,29 +97,28 @@ jobs:
       - name: "Build Python release artifacts"
         run: |
           cd cpython
-          python ../release.py --export ${{ inputs.cpython_release }} --skip-docs
```
sethmlarson, line 73: (also here)

.github/workflows/source-and-docs-release.yml
```diff
@@ -135,10 +161,10 @@ jobs:
       - name: "Test Python source tarballs"
         run: |
           mkdir -p ./tmp/installation/
-          cp Python-${{ inputs.cpython_release }}.tgz ./tmp/
+          cp Python-${{ env.CPYTHON_RELEASE }}.tgz ./tmp/
```
sethmlarson, line 164: (also here)

.github/workflows/source-and-docs-release.yml
```diff
@@ -18,27 +36,36 @@ on:
...
-          ref: "v${{ inputs.cpython_release }}"
+          repository: "${{ env.GIT_REMOTE }}/cpython"
+          ref: "v${{ env.CPYTHON_RELEASE }}"
           path: "cpython"
 
       - name: "Verify CPython commit matches tag"
         run: |
-          if [[ "${{ inputs.git_commit }}" != "$(cd cpython && git rev-parse HEAD)" ]]; then
-            echo "expected git commit ('${{ inputs.git_commit }}') didn't match tagged commit ('$(git rev-parse HEAD)')"
+          if [[ "${{ env.GIT_COMMIT }}" != "$(cd cpython && git rev-parse HEAD)" ]]; then
```
sethmlarson, line 67: Since we're using environment variables now instead of GitHub Action expressions, can we replace the "${{ NAME }}" with "$NAME" when we're running commands?

sethmlarson (commented):
This is a good direction! I had one comment regarding extra hardening against script injection if we're using environment variables now.

hugovk:
Interesting, I didn't get these shellcheck errors locally:

```
.github/workflows/source-and-docs-release.yml:98:9: shellcheck reported issue in this script: SC2086:info:2:31: Double quote to prevent globbing and word splitting [shellcheck]
```

.github/workflows/source-and-docs-release.yml
```diff
@@ -18,27 +36,36 @@ on:
...
+          repository: "${{ env.GIT_REMOTE }}/cpython"
+          ref: "v${{ env.CPYTHON_RELEASE }}"
           path: "cpython"
 
       - name: "Verify CPython commit matches tag"
         run: |
-          if [[ "${{ inputs.git_commit }}" != "$(cd cpython && git rev-parse HEAD)" ]]; then
-            echo "expected git commit ('${{ inputs.git_commit }}') didn't match tagged commit ('$(git rev-parse HEAD)')"
+          if [[ "$GIT_COMMIT" != "$(cd cpython && git rev-parse HEAD)" ]]; then
+            echo "expected git commit ('$GIT_COMMIT') didn't match tagged commit ('$(git rev-parse HEAD)')"
```
hugovk, line 68: This one failed with a manual trigger, will have to check it later:

https://github.com/hugovk/release-tools/actions/runs/9389770445/job/25858031818

hugovk:
Ready for review.

Here's the latest manual trigger: https://github.com/hugovk/release-tools/actions/runs/9400901728

sethmlarson (approved):
LGTM, thanks Hugo!

PR #137, opened 2024-06-17
Add type hints to `run_release.py`

hugovk:
For https://github.com/python/release-tools/issues/114.

If temporarily removing `run_release.py` from the mypy exclude list, this fixes 57% of the errors:

```
Found 133 errors in 3 files (checked 6 source files)
```
->
```
Found 76 errors in 3 files (checked 6 source files)
```

### DIFF
--- a/mypy-requirements.txt
+++ b/mypy-requirements.txt
@@ -1,5 +1,5 @@
-aiohttp
+aiohttp==3.9.4
 mypy==1.10
 pytest
 pytest-mock
-sigstore
+sigstore==1.1.2
--- a/mypy-requirements.txt
+++ b/mypy-requirements.txt
@@ -1,3 +1,5 @@
+aiohttp
 mypy==1.10
 pytest
 pytest-mock
+sigstore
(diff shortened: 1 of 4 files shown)


run_release.py
```diff
@@ -926,8 +929,6 @@ def modify_the_release_to_the_prerelease_pages(db: DbfilenameShelf) -> None:
 
 
 def post_release_merge(db: DbfilenameShelf) -> None:
-    release_tag: release_mod.Tag = db["release"]
```
hugovk, line 929: This was a duplicate of line 936 below.

mypy-requirements.txt
```diff
@@ -1,3 +1,5 @@
+aiohttp
```
JelleZijlstra, line 1: Should we pin versions here?

JelleZijlstra (approved):

mypy-requirements.txt
hugovk: Sure, updated!

sethmlarson (approved):

PR #140, opened 2024-06-27
Update `run_release.py`

hugovk:
Some smallish updates and fixes:

* Early on, we check the `docker` executable is available, this is good. But it failed for me later on because Docker wasn't running. Add an early check that it's running by calling `docker container ls`.

* When running on macOS, it shows a warning about running on non-Linux. Let's move this after the args have been parsed, so I can check `--help` without the warning.

* Fix argparse description: we're making a CPython release, not (only) processing some integers.

* Update a task description: it checks both downloads and docs servers.

* The bits in `API_KEY_REGEXP` are "user" and "key", not "major" and "minor" (probably copy/pasted from `RELEASE_REGEXP`).

* Fix some typos and capitalisation.

### DIFF
--- a/run_release.py
+++ b/run_release.py
@@ -1154,7 +1154,7 @@ WARNING! This script has not been tested on a platform other than Linux.
 
 Although it should work correctly as long as you have all the dependencies,
 some things may not work as expected. As a release manager, you should try to
-fix these things in this script so it also support your platform.
+fix these things in this script so it also supports your platform.
 """
         )
         if not ask_question("Do you want to continue?"):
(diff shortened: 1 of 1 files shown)


run_release.py
```diff
@@ -1154,28 +1143,49 @@ def _api_key(api_key: str) -> str:
...
     assert isinstance(auth_key, str), "We need an AUTH_INFO env var or --auth-key"
+
+    if "linux" not in sys.platform:
+        print(
+            """\
+WARNING! This script has not been tested on a platform other than Linux.
+
+Although it should work correctly as long as you have all the dependencies,
+some things may not work as expected. As a release manager, you should try to
+fix these things in this script so it also support your platform.
```
JelleZijlstra, line 1157:
```suggestion
fix these things in this script so it also supports your platform.
```

sethmlarson (approved):

PR #147, opened 2024-07-23
Remove space from '32 -bit'

hugovk:
Re: https://discuss.python.org/t/why-is-there-an-extra-space-in-download-windows-installer-32-bit/46589.

Since https://github.com/python/release-tools/commit/819af73e38603262ae80b718535a6fabfa35386f, there's been a rogue space in "Windows installer (32 -bit)" on download pages, for example:

https://www.python.org/downloads/release/python-3124/

This will fix future uploads. Ee can then do a query to fix old ones.

### DIFF
--- a/add-to-pydotorg.py
+++ b/add-to-pydotorg.py
@@ -116,7 +116,7 @@ def get_file_descriptions(release):
             rx(r"-embed-win32\.zip$"),
             ("Windows embeddable package (32-bit)", 1, False, ""),
         ),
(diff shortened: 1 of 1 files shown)


ewdurbin:
```
app=> begin;
BEGIN
app=*> update downloads_releasefile set name='Windows installer (32-bit)' where name ='Windows installer (32 -bit)';
UPDATE 39
app=*> commit;
```

PR #148, opened 2024-07-24
Remove unused CHM handling

hugovk:
We stopped distributing ``.chm`` (Windows HTML help) files in Python 3.11, see https://github.com/python/cpython/issues/91242.

And Python 3.8-3.11 are in security mode, meaning only the source is released, no CHM files.

For example:

* https://www.python.org/downloads/release/python-3124/ - 3.12: installers but no CHM
* https://www.python.org/downloads/release/python-3819/ - 3.8: source only and no CHM

We can now remove some redundant CHM code.

### DIFF
--- a/add_to_pydotorg.py
+++ b/add_to_pydotorg.py
@@ -122,7 +122,6 @@ def get_file_descriptions(
             ("Windows embeddable package (32-bit)", 1, False, ""),
         ),
         (rx(r"\.exe$"), ("Windows installer (32-bit)", 1, v < (3, 9), "")),
-        (rx(r"\.chm$"), ("Windows help file", 1, False, "")),
         (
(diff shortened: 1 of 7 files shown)


zooba:
@ambv Did this go in before the release last week? Or does it still need a test run?

hugovk:
It went in before the 3.13 RC1 release.

This PR was merged in the afternoon (Europe time) on 31st July and the release build was later that night: https://dev.azure.com/Python/cpython/_build/results?buildId=154882&view=results

And that build shows it used commit https://github.com/python/release-tools/commit/fe9236f75be51818f36fcfee05d1b7cd84d626cc, which came after this merge: https://github.com/python/release-tools/commits/master/

PR #150, opened 2024-07-27
Add `--no-gpg` to skip signing with GPG

hugovk:
@sethmlarson and I are planning on replacing GPG signing with Sigstore for 3.14.

Here's a first step to update the tooling for this, adding a `--no-gpg` flag to skip signing with GPG. By default, GPG signing is still used, per the the status quo.

(Also fix a "django"->"Django" typo.)

### DIFF
--- a/release.py
+++ b/release.py
@@ -580,7 +580,7 @@ def upload(tag: Tag, username: str) -> None:
         )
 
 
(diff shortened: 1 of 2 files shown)


sethmlarson (approved):
LGTM, thanks Hugo!

PR #151, opened 2024-07-31
Add types with enforcement to SBOM

ambv:

PR #152, opened 2024-07-31
Enforce typing in run_release.py and release.py

### LINKED ISSUE #114: Add a type checker

ambv:
This depends on #151 landing first. (change base branch to `master` to land)

Closes #114.

### DIFF
--- a/release.py
+++ b/release.py
@@ -37,7 +37,7 @@ SPACE = " "
 tag_cre = re.compile(r"(\d+)(?:\.(\d+)(?:\.(\d+))?)?(?:([ab]|rc)(\d+))?$")
 
 
-class ReleaseShelf(Protocol):  # this will be easier to type in 3.13
+class ReleaseShelf(Protocol):
     def close(self) -> None: ...
 
     @overload
--- a/release.py
+++ b/release.py
@@ -20,14 +20,110 @@ import subprocess
 import sys
 import tempfile
 from contextlib import contextmanager
-from typing import Any, Callable, Generator, Self
+from dataclasses import dataclass
+from pathlib import Path
+from typing import (
+    Any,
+    Callable,
+    Generator,
+    Literal,
+    Protocol,
+    Self,
+    overload,
+)
 
 COMMASPACE = ", "
 SPACE = " "
 tag_cre = re.compile(r"(\d+)(?:\.(\d+)(?:\.(\d+))?)?(?:([ab]|rc)(\d+))?$")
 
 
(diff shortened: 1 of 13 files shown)


release.py
```diff
@@ -20,14 +20,110 @@
...
+    overload,
+)
 
 COMMASPACE = ", "
 SPACE = " "
 tag_cre = re.compile(r"(\d+)(?:\.(\d+)(?:\.(\d+))?)?(?:([ab]|rc)(\d+))?$")
 
 
-# Ideas stolen from Mailman's release script, Lib/tokens.py and welease
+class ReleaseShelf(Protocol):  # this will be easier to type in 3.13
```
ambv, line 40: This is like a `TypedDict` but for a pseudo-mapping not inheriting from a `dict`. The syntax is a bit horrible but it's per-key strongly typed.

And actually, this *won't* be easier in 3.13 :sob:
https://peps.python.org/pep-0705/#a-typedmapping-protocol-type

run_release.py
```diff
@@ -652,25 +648,19 @@ def execute_command(command: str) -> None:
 
     # Docs
 
-    release_tag: release_mod.Tag = db["release"]
+    release_tag = db["release"]
     if release_tag.is_final or release_tag.is_release_candidate:
         source = f"/home/psf-users/{db['ssh_user']}/{db['release']}"
         destination = f"/srv/www.python.org/ftp/python/doc/{release_tag}"
 
-        def execute_command(command: str) -> None:
```
ambv, line 660: This is shadowing an identical function from L641

hugovk:
#151 has landed, I resolved the conflicts.

release.py
hugovk, line 40: Then remove the comment? 
```suggestion
class ReleaseShelf(Protocol):
```

hugovk (approved):

run_release.py
```diff
@@ -179,15 +175,6 @@
 """
 
 
-@dataclass
-class Task:
```
sethmlarson, line 183: `Task` isn't used anywhere in `release.py` that I can see, is there a reason we moved it there?

sethmlarson (approved):
Only had one comment, otherwise LGTM!

run_release.py
ambv: Yeah, it's used in typing of ReleaseShelf and it refers to ReleaseShelf itself. So it needs to be in the same spot.

PR #153, opened 2024-07-31
Check for magic number mismatch early to avoid surprises in RC stage

ambv:
In RC/final stage this breaks the build. In previous stages it generates a question whether to continue.

This depends on #152. After that lands, change the base branch here to master. Then you can merge the PR.

### DIFF
--- a/tests/test_run_release.py
+++ b/tests/test_run_release.py
@@ -1,6 +1,10 @@
+from pathlib import Path
+from typing import cast
+
 import pytest
 
 import run_release
+from release import ReleaseShelf, Tag
 
 
 @pytest.mark.parametrize(
@@ -22,3 +26,14 @@ def test_invalid_extract_github_owner() -> None:
         "https://example.com",
     ):
         run_release.extract_github_owner("https://example.com")
+
+
+def test_check_magic_number() -> None:
+    db = {
+        "release": Tag("3.13.0rc1"),
+        "git_repo": str(Path(__file__).parent / "magicdata"),
+    }
+    with pytest.raises(
+        run_release.ReleaseException, match="Magic numbers in .* don't match"
+    ):
+        run_release.check_magic_number(cast(ReleaseShelf, db))
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,6 @@ dist/
 downloads/
 eggs/
 .eggs/
-lib/
 lib64/
 parts/
 sdist/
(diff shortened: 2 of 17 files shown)


ambv:
Wow, that codecov sure is a snitch.

tests/test_run_release.py
```diff
@@ -22,3 +26,14 @@ def test_invalid_extract_github_owner() -> None:
         "https://example.com",
     ):
         run_release.extract_github_owner("https://example.com")
+
+
+def test_check_magic_number() -> None:
+    db = {
+        "release": Tag("3.13.0rc1"),
```
hugovk, line 33: We could also test that non-RC/non-final does not raise.

hugovk (approved):

tests/test_run_release.py
ambv: It's a little annoying because it asks a question so we'd have to mock that out. I'll mark that as "future improvement".

Yhg1s:
Doh, this broke releases with older versions, which don't have pycore_magic_number.h:
```
💥  Checking the magic number is up-to-date
Traceback (most recent call last):
  File "./run_release.py", line 1283, in <module>
    main()
  File "./run_release.py", line 1279, in main
    automata.run()
  File "./run_release.py", line 244, in run
    raise e from None
  File "./run_release.py", line 241, in run
    self.current_task(self.db)
  File "release.py", line 126, in __call__
    return getattr(self, "function")(db)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./run_release.py", line 402, in check_magic_number
    magic_actual = get_magic(magic_actual_file, magic_actual_re)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./run_release.py", line 391, in get_magic
    if m := regex.search(source.read_text()):
                         ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/pathlib.py", line 1059, in read_text
    with self.open(mode='r', encoding=encoding, errors=errors) as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/pathlib.py", line 1045, in open
    return io.open(self, mode, buffering, encoding, errors, newline)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'cpython/Include/internal/pycore_magic_number.h'
```

PR #159, opened 2024-08-08
Upgrade to Sigstore CLI v3

sethmlarson:
Part of https://github.com/python/cpython/issues/122785 cc @woodruffw


add_to_pydotorg.py
```diff
@@ -340,14 +340,36 @@ def has_sigstore_signature(filename: str) -> bool:
...
+                    "--bundle",
+                    bundle_file,
+                    filename,
+                ]
+            )
+
+            # New sigstore CLI makes Sigstore bundles end in '.json'
+            # but our users expect bundles to be '.sigstore'
+            if os.path.exists(bundle_file + ".json"):
+                os.rename(bundle_file + ".json", bundle_file)
```
woodruffw, line 368: I might be wrong but this should never happen, right? If you're explicitly passing `--bundle {bundle_file}` above, then sigstore-python should never emit the newer `.sigstore.json` suffix, only whatever filename you explicitly pass 🙂

sethmlarson: Done in https://github.com/python/release-tools/pull/159/commits/4f724be99dd83c58f8dd0ad63956ca912c7de8c3

sethmlarson:
After this is merged, we'll need to upgrade the dependencies on the python.org server as well. That doesn't appear to be a part of the release process, so wondering now if that should be added to ensure we're using a consistent set over there, too.

hugovk (approved):

PR #162, opened 2024-08-12
Recursively purge all download URLs

### LINKED ISSUE #156: CDN purges don't purge all files.
Currently, we purge the CDN cache for a couple of files, but not all the files (notably missing are new files added relatively recently: sboms and sigstore signatures). https://github.com/python/release-tools/blob/master/run_release.py#L931

We should update that list, or just list all files on the download server and purge all of them.

sethmlarson:
Closes #156, this implementation recurses into sub-directories to find all artifacts instead of relying on a pre-defined set.


hugovk:
This traverses directories on the FTP and purges files it finds. 

But when things need purging, will these things be discoverable if they haven't been purged yet?

PR #164, opened 2024-08-16
cpython#121277: Replace `next` versions in docs by the just-released version

encukou:
See CPython issue: https://github.com/python/cpython/issues/121277

After the CPython PR is merged, a Sphinx plugin will expand `next` in directives like `versionchanged` to e.g. `3.14.0b0 (unreleased)`.

This PR gives the RM a tool to replace all such occurences of `next` by the currently released version, baking it directly into the ReST source files, as part of the version bump commit.

This is done textually (I don't know of a round-trip parser for ReST).

To make sure the text manipulation worked, the released docs artifact (if it's being published) is grepped for the string `(unreleased)`, and if it's found, the RM is asked for confirmation.


encukou:
Dear RMs, do you want this?

I can add tests for the remaining uncovered lines; is this the style of tests you're looking for?

update_version_next.py
```diff
@@ -0,0 +1,95 @@
...
+        )
+        next
+        (?P<after>
+            .*
+        )
+    """,
+    re.VERBOSE | re.DOTALL,
+)
+
+doc_dir = Path(__file__).parent.parent.resolve()  # cpython/Doc/tools  # cpython/Doc
```
hugovk, line 35: Black has jumbled up these comments:

```python
doc_dir = (Path(__file__)
           .parent  # cpython/Doc/tools
           .parent  # cpython/Doc
           .resolve()
           )
```

[`c482123` (#164)](https://github.com/python/release-tools/pull/164/commits/c482123fe488581ea1f8817935a02eeaf353cda8#diff-ae24a10c93bce91291f30c5c206ef14058dc8d4f8dc73c36cbf9770112a8e1ddL35)

This worked if this tool was in the CPython repo (where it was originally, in https://github.com/python/cpython/pull/121278) but maybe we don't want a default here?

`run_release.py` doesn't have a default for `--repository REPO    Location of the CPython repository` and in practice builds I've been calling it as `--repository ../cpython` but it could be anywhere.

update_version_next.py
```diff
@@ -0,0 +1,95 @@
...
+
+parser = argparse.ArgumentParser(
+    description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
+)
+parser.add_argument(
+    "version",
+    help='String to replace "next" with. Usually `x.y`, ' + "but can be anything.",
```
hugovk, line 42:
```suggestion
    help='String to replace "next" with. Usually `x.y`, but can be anything.',
```

tests/test_update_version_next.py
```diff
@@ -0,0 +1,90 @@
...
+
+.. VERSIONCHANGED: next
+
+.. deprecated-removed: 3.0 next
+"""
+
+EXPECTED_CHANGED = TO_CHANGE.replace("next", "VER")
+
+
+class TestVersionNext(unittest.TestCase):
```
hugovk, line 65: We can use pytest-style tests here with class-less functions and plain asserts.

hugovk:
> Dear RMs, do you want this?

Yes, I think so, but still need to take a closer look, and would also like to hear from other RMs who have actually made releases :)

> I can add tests for the remaining uncovered lines; is this the style of tests you're looking for?

I've not had a proper look yet, but will do.

tests/test_update_version_next.py
encukou: And magic fixtures, right? :)

hugovk (commented):
Looking good, I tested it against https://github.com/python/cpython/pull/121278 and it worked as expected.

Here's a few small suggestions.

update_version_next.py
```diff
@@ -0,0 +1,92 @@
...
+    "-v",
+    action="count",
+    default=0,
+    help="Increase verbosity. Can be repeated (`-vv`).",
+)
+
+
+def main(argv: list[str]) -> None:
+    args = parser.parse_args(argv)
+    version = args.version
```
hugovk, line 59: Shall we validate that the input is of the form "x.y"?

encukou: I'd just trust the caller here.

run_release.py
```diff
@@ -497,6 +498,13 @@ def bump_version(db: ReleaseShelf) -> None:
     )
 
 
+def bump_version_in_docs(db: ReleaseShelf) -> None:
+    update_version_next.main([db["release"].doc_version, str(db["git_repo"])])
+    subprocess.check_call(
+        ["git", "commit", "-a", "--amend", "--no-edit"], cwd=db["git_repo"]
```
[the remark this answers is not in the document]
encukou: All the other `["git", "commit"…` use `"-a"`, I'd leave it here too for consistency.

update_version_next.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
encukou: I made it executable.

hugovk (approved):
Thanks!

encukou:
On the sprint, @Yhg1s agreed to the concept, so I'll merge.

I'll link related PRs on the CPython issue: https://github.com/python/cpython/issues/121277

PR #168, opened 2024-09-10
Drop letter PDF: don't wait for letter PDF zips

hugovk:
First step for https://github.com/python/docs-community/issues/128.

Instead of building both A4 and letter PDFs, we're only to build A4. This will significantly reduce the docs build times.


willingc (approved):
Nice! Saving ⌛

AA-Turner (approved):

sethmlarson (approved):

PR #170, opened 2024-09-25
Add issue templates to files to change

zware:
...when bumping the minor version number.

Also fix a minor typo in the area.

### DIFF
--- a/release.py
+++ b/release.py
@@ -424,9 +424,7 @@ def bump(tag: Tag) -> None:
     tweak_patchlevel(tag)
 
     extra_work = False
-    # Older releases have a plain text README,
-    # newer releases have README.rst.
-    other_files = ["README.rst", "README", "Misc/NEWS"]
+    other_files = ["README.rst"]
     if tag.patch == 0 and tag.level == "a" and tag.serial == 0:
         extra_work = True
         other_files += [
--- a/release.py
+++ b/release.py
@@ -438,6 +438,8 @@ def bump(tag: Tag) -> None:
             "Doc/license.rst",
             "PC/pyconfig.h.in",
             "PCbuild/rt.bat",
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):

zware:
The goal here is to avoid the situation where for some months we offered the option of 3.13 in the `bug.yml` template (thanks to python/cpython#111083) but not the `crash.yml` template (finally added in python/cpython/#124560.  Ideally it would be noted somewhere that it's actually the *previous* version being added to these files when this happens, but I think context should make that clear enough to our RMs.

zware:
(Side note: why am I allowed to merge here? I would have thought this should be limited to RMs :))

release.py
```diff
@@ -438,6 +438,8 @@ def bump(tag: Tag) -> None:
             "Doc/license.rst",
             "PC/pyconfig.h.in",
             "PCbuild/rt.bat",
+            ".github/ISSUE_TEMPLATE/bug.yml",
+            ".github/ISSUE_TEMPLATE/crash.yml",
```
hugovk, line 442: Please could you sort this whole list?

hugovk (approved):
Thanks, please can you also add them to PEP 101?

release.py
zware: One could argue that it's currently sorted logically rather than alphabetically and beyond that the only thing out of order is `LICENSE` before `Doc/license.rst` (if you consider `.` to sort to the end).  Happy to sort as you want, but I don't want to pick the wrong sort :)

zware:
See python/peps#4003 for the PEP-101 update.

release.py
hugovk: Good point, let's just leave it like this. They'll be popped open into an editor so it's helpful to have similar ones grouped :)

hugovk (approved):
Thanks!

PR #172, opened 2024-10-07
Enable use of Azure Trusted Signer for code signing

zooba:
This isn't quite ready to merge - I'm hoping we can get Nuget signing sorted first. But if not then I'll remove the certificate requirement from upload and we'll just go back to unsigned packages (the contents are still signed).

Everything else will sign with our new Azure Trusted Signer account.

I need to update the readme.


zooba:
So the story on Nuget signing is that it's possible with a new .NET sign tool, but that tool insists on opening the package and re-signing everything inside. It does this for MSIX packages as well, which actually leads to a failure because it signs files that it shouldn't be trying to sign.

I've [requested](https://github.com/dotnet/sign/issues/775) an option to bypass this, which would make it a totally suitable tool. Until then, it seems fine to skip NuGet packages for the time being (NuGet [can't validate](https://github.com/NuGet/Home/pull/13791) Azure Trusted Signing packages anyway).

zooba:
Hopefully the build at https://dev.azure.com/Python/cpython/_build/results?buildId=156601&view=results will succeed (without publishing), and if so then the build changes are good. Just need to update readme docs.

hugovk (approved):
That build is green. Would you like to merge this before today's 3.14 alpha 1?

zooba:
I'm running one more full test run (including PGO) to make sure it all works with the current tip, and then yes.

PR #178, opened 2024-10-19
Use sentence case for headers in 'What's new' template

hugovk:
Re: https://devguide.python.org/documentation/style-guide/#capitalization

Already updated for 3.14 in https://github.com/python/cpython/pull/125438, let's update the template for the next branch.


run_release.py
```diff
@@ -44,7 +44,7 @@
 
 WHATS_NEW_TEMPLATE = """
 ****************************
-  What's New In Python {version}
+  What's new in Python {version}
```
AA-Turner, line 47: Appreciate it is an exception to the rule, but *What's New* as the name of a series of articles seems better capitalised.
```suggestion
  What's New In Python {version}
```

run_release.py
```diff
@@ -111,40 +111,58 @@
...
 
 Deprecated
 ==========
 
+* module_name:
+  TODO
+
 
 
+.. Add deprecations above alphabetically, not here at the end.
```
AA-Turner, line 158:
```suggestion
.. Add deprecations above alphabetically, not here at the end.

```

run_release.py
```diff
@@ -154,22 +172,28 @@
...
 ----------------------
 
+* TODO
+
 Deprecated
 ----------
 
+* TODO
+
+.. Add C API deprecations above alphabetically, not here at the end.
```
AA-Turner, line 196:
```suggestion
.. Add C API deprecations above alphabetically, not here at the end.

```

run_release.py
```diff
@@ -154,22 +172,28 @@
...
 Removed
 -------
```
AA-Turner, line 198: A unique title means that it will have an `#anchor` instead of `#id<000>`
```suggestion
Removed C APIs
--------------
```

run_release.py
```diff
@@ -154,22 +172,28 @@
...
 ------------
 
+* TODO
 Porting to Python {version}
 ----------------------
 
+* TODO
+
 Deprecated
 ----------
```
AA-Turner, line 192: A unique title means that it will have an `#anchor` instead of `#id<000>`
```suggestion
Deprecated C APIs
-----------------
```

AA-Turner (commented):
It may also be worth moving some of the sections around - but I could suggest that in a follow-up PR (removals above deprecations, combining the two "porting", etc)

run_release.py
hugovk: It's still a title and I'd prefer keeping sentence case. This also matches the frontpage: https://docs.python.org/3/

run_release.py
```diff
@@ -100,7 +100,7 @@
    so it's worth checking back even after reading earlier versions.
 
 
-Summary -- Release highlights
+Summary -- release highlights
 =============================
```
ezio-melotti, line 104:
```suggestion
Summary --- release highlights
==============================
```
Given the dash, this could also work with an uppercase `R`, but up to you:
```suggestion
Summary --- Release highlights
==============================
```

PR #181, opened 2024-10-23
Synchronize the latest CPython repo changes with `purge.py`

Wulian233:
At https://github.com/python/cpython/commit/da1e5526aee674bb33c17a498aa3781587b9850c
Remove these lines

### DIFF
--- a/windows-release/purge.py
+++ b/windows-release/purge.py
@@ -53,7 +53,6 @@ FILES = [
     "test_pdb.msi",
     "tools.msi",
     "ucrt.msi",
-    "Windows6.0-KB2999226-x64.msu",
 ]
 PATHS = [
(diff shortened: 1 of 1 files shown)


Wulian233:
I don't know if this file for the cpython repository is valid for the repository itself. The purge.py repository implemented free threading 9 months ago. 
https://github.com/python/release-tools/commit/76e1f69891816cca6feb11840ccf5b79b341be4f

Without synchronizing the CPython repository (if this file from the cpython repository is useless, can we delete it?).

zooba:
It's no big deal either way. There's no harm in purging files that don't exist from the CDN cache, and these MSUs aren't coming back (they're included by default in all Win10). And it's rare these days that we need to do a purge, as our builds are pretty reliable.

PR #184, opened 2024-10-25
Don't show warning when releasing from macOS

hugovk:
Python 3.14.0a1 was released using macOS, let's remove the warning on macOS.


hugovk:
a2 too

PR #187, opened 2024-10-30
More test coverage for release.py and sbom.py

hugovk:
* `release.py`: 43% -> 51% coverage
* `sbom.py`: 49% -> 52% coverage

### DIFF
--- a/release.py
+++ b/release.py
@@ -391,8 +391,10 @@ def constant_replace(
     os.rename(fn + ".new", fn)
 
 
-def tweak_patchlevel(tag: Tag, done: bool = False) -> None:
-    print("Updating Include/patchlevel.h...", end=" ")
+def tweak_patchlevel(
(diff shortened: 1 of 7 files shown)


release.py
```diff
@@ -391,8 +391,10 @@ def constant_replace(
     os.rename(fn + ".new", fn)
 
 
-def tweak_patchlevel(tag: Tag, done: bool = False) -> None:
-    print("Updating Include/patchlevel.h...", end=" ")
+def tweak_patchlevel(
+    tag: Tag, fn: str = "Include/patchlevel.h", done: bool = False
```
sethmlarson, line 395: Can we use a more descriptive parameter name? I read `fn` as "function" at first and was confused.

sethmlarson (approved):
LGTM, only had one nit.

release.py
hugovk: Sure, I copied `fn` from elsewhere in the file, so updated them all to `filename`.

PR #189, opened 2024-11-06
PEP 761: Python 3.14+ discontinues PGP signatures

hugovk:
[PEP 761](https://peps.python.org/pep-0761/) has been accepted:

* https://discuss.python.org/t/pep-761-deprecating-pgp-signatures-for-cpython-artifacts/67180/39
* https://github.com/python/steering-council/issues/260

This PEP automatically skips PGP signatures for 3.14+.

It also removes the `--no-gpg` flag because 3.13 and older will still be PGP signed.

cc @sethmlarson

### DIFF
--- a/run_release.py
+++ b/run_release.py
@@ -1264,7 +1264,8 @@ fix these things in this script so it also supports your platform.
                 "This release script is not compatible with the running platform"
             )
 
-    no_gpg = args.no_gpg
+    release_tag = release_mod.Tag(args.release)
+    no_gpg = args.no_gpg or release_tag.as_tuple() >= (3, 14)
     tasks = [
         Task(check_git, "Checking Git is available"),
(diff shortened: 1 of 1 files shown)


run_release.py
```diff
@@ -1264,7 +1259,8 @@ def _api_key(api_key: str) -> str:
                 "This release script is not compatible with the running platform"
             )
 
-    no_gpg = args.no_gpg
+    release_tag = release_mod.Tag(args.release)
+    no_gpg = release_tag.as_tuple() >= (3, 14)
```
JelleZijlstra, line 1263:
```suggestion
    no_gpg = release_tag.as_tuple() >= (3, 14)  # see PEP 761
```

sethmlarson (approved):
:partying_face: Thanks Hugo!

PR #190, opened 2024-11-07
Switch to new signing tool for Windows releases

zooba:
The dotnet sign tool is capable of signing all our file types that we need using Trusted Signing, which should simplify our use of it.

### DIFF
--- a/windows-release/azure-pipelines.yml
+++ b/windows-release/azure-pipelines.yml
@@ -165,9 +165,8 @@ stages:
     jobs:
     - template: stage-pack-nuget.yml
       parameters:
-        # Nuget signing is disabled because Azure Trusted Signing does not support it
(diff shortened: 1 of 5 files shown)


zooba:
We can't swap over yet due to a few more Nuget-related limitations.

zooba:
It may not be a limitation - Nuget doesn't like signing with the lifetime signing EKU[^1] attached to the test cert. So it's possible that real signing will work fine. I'm following up with the relevant teams to see what the best approach here is.

We still don't have to hurry to merge this change. Nuget.org doesn't validate Azure Trusted Signing signatures yet anyway.

[^1]: A marker to indicate that signatures shouldn't be trusted beyond the lifetime of the certificate, even if timestamped.

zooba:
Signing works fine with the real certificate (tested, but deleted the test build to avoid having extra signed builds of Python floating around). So this is okay to merge.

@sethmlarson, don't suppose you're interested in taking a look?

PR #192, opened 2024-11-19
Make Nuget signing optional (disabled by default)

zooba:
Nuget will not allow uploading signed packages unless the certificate is registered, and we cannot register Azure Trusted Signing certificates at this time. So it's best to just not sign the package.


zooba:
Creating a build with and without the option set succeeds, which ought to be enough testing for this change.

PR #201, opened 2024-12-03
Make SPDX IDs stable for non-collisions

sethmlarson:
By not keeping IDs stable it meant that we weren't matching our source SBOM's IDs, specifically `SPDXRef-PACKAGE-cpython`. This keeps the IDs stable unless there is a collision, which can only happen w/ filepaths not package names.

### DIFF
--- a/tests/test_sbom.py
+++ b/tests/test_sbom.py
@@ -14,11 +14,11 @@ import sbom
 @pytest.mark.parametrize(
     ["value", "expected"],
     [
-        ("abc", "abc-ba7816bf"),
-        ("def", "def-cb8379ac"),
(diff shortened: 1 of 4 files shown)


tests/test_sbom.py
```diff
@@ -14,11 +14,11 @@
 @pytest.mark.parametrize(
     ["value", "expected"],
     [
-        ("abc", "abc-ba7816bf"),
```
Yhg1s, line 17: As far as I can tell this doesn't actually exercise the unhappy hash path, would it make sense to add a collision to this set of parameters? I don't remember if pytest guarantees ordered execution of the test with parameters.

Yhg1s (approved):

PR #202, opened 2024-12-04
Add the pip ecosystem to dependabot.yml

ezio-melotti:
I noticed that the `pip` deps are not being updated automatically.
This PR adds the `pip` ecosystem to the `dependabot.yml`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,6 +13,8 @@ updates:
     directory: "/"
     schedule:
       interval: monthly
+    assignees:
+      - "ezio-melotti"
     groups:
       pip:
         patterns:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,3 +8,12 @@ updates:
       actions:
         patterns:
           - "*"
+
+  - package-ecosystem: pip
+    directory: "/"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -8,3 +8,12 @@ updates:
       actions:
         patterns:
           - "*"
+
+  - package-ecosystem: pip
+    directory: "/"
+    schedule:
+      interval: monthly
```
ezio-melotti, line 15:
```suggestion
    schedule:
      interval: monthly
    assignees:
      - "ezio-melotti"
```
If would like me to review (and possibly merge) these PRs, I don't mind have them auto-assigned to me.

hugovk (approved):
This is a good idea, note that we need to be careful because pip-compile has a bug (https://github.com/jazzband/pip-tools/pull/2082) that breaks the ordering:

```diff
diff --git a/requirements.txt b/requirements.txt
---only-binary :all:
 --no-binary grapheme
 --no-binary grpclib
+--only-binary :all:
```

So until they have a new release (https://github.com/jazzband/pip-tools/issues/2112) we'll need to manually fix that up in Dependabot PRs before merge.

PR #205, opened 2024-12-05
Test Python 3.13 and use 3.12 to build sdist and docs

hugovk:
I added 3.13 to tox and the CI, then ran `tox -e py313` and for each package that needed upgrading for Python 3.13, ran something like:

`pip-compile --generate-hashes --output-file=requirements.txt requirements.in --upgrade-package aiohttp`

I also used pip-tools on Python 3.12 because we have 3.12 on the downloads and docs servers.

Then also bumped to use Python 3.12 to build the sdist and docs on the CI.

---

I didn't add 3.14 for testing yet, because we want to use `--only-binary` and there's some things which don't have 3.14 binaries yet (coverage was the first).


hugovk:
@sethmlarson How do these `pip-compile` updates look?

hugovk:
To resolve the conflict, I deleted `requirements.txt`, ran `pip-compile --generate-hashes --output-file=requirements.txt requirements.in` and manually fixed the `--only-binary :all:` bug (https://github.com/jazzband/pip-tools/pull/2082).

Then upgraded `aiohttp` and `attrs` in `dev-requirements.txt` so it matches those in `requirements.txt`.

sethmlarson (approved):
LGTM, thanks Hugo!

PR #206, opened 2024-12-11
Add zizmor security linter to pre-commit

hugovk:
> `zizmor` is a static analysis tool for GitHub Actions. It can find many common security issues in typical GitHub Actions CI/CD setups.

https://woodruffw.github.io/zizmor/

Seth also recommends avoiding using shared caches in the build and release process:

https://blog.pypi.org/posts/2024-12-11-ultralytics-attack-analysis/

Reviewing the workflows here, we use caches in https://github.com/python/release-tools/blob/master/.github/workflows/test.yml to run unit tests and https://github.com/python/release-tools/blob/master/.github/workflows/lint.yml to lint the code with pre-commit, but importantly don't use any in https://github.com/python/release-tools/blob/master/.github/workflows/source-and-docs-release.yml that creates release artifacts.

I also updated the other pre-commit hooks including new typing and f-string upgrades.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.7.1
+    rev: v0.8.2
     hooks:
       - id: ruff
         args: [--exit-non-zero-on-fix]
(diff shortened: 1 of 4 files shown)


sethmlarson (approved):
LGTM

PR #207, opened 2024-12-15
Remove unused `buildbotapi` code and test the rest

hugovk:
About half of `buildbotapi.py` is unused code.

Let's remove it, and add tests with 100% coverage for the remaining code.


hugovk:
The next 3.14 alpha is tomorrow, if there's no objections I'll merge this tomorrow morning before starting the release so I'll be the first to test it.

PR #214, opened 2025-02-10
Disambiguate SPDXIDs when merging source and externals SBOMs

sethmlarson:
Closes https://github.com/python/cpython/issues/129706


hugovk:
CI is failing:
* lint because of missing type hints
* build because of stricter blurb, should be fixed by https://github.com/python/release-tools/pull/215

Would you like this in tomorrow's 3.14 alpha?

sethmlarson:
> Would you like this in tomorrow's 3.14 alpha?

I don't think it's a blocker, especially for an alpha. Thanks for the quick review, I'll get your comments addressed.

hugovk (approved):

hugovk:
It's ready in time for the alpha, let's merge, thanks!

sethmlarson:
Thanks Hugo! :purple_heart:

PR #224, opened 2025-03-16
Automatically bump the README version

hugovk:
Add a function `tweak_readme` to open `README.rst`, update the first line to (for example) "This is Python version 3.14.0 alpha 7" and update the underline to be the same length.

Also add some tests for `Tag.is_final`.

### DIFF
--- a/release.py
+++ b/release.py
@@ -439,7 +439,7 @@ def tweak_readme(tag: Tag, filename: str = "README.rst") -> None:
     # and update length of underline in second line to match.
     lines = readme.read_text().splitlines()
     this_is = f"This is Python version {tag.long_name}"
-    underline = f"{'=' * len(this_is)}"
+    underline = "=" * len(this_is)
     lines[0] = this_is
     lines[1] = underline
 
--- a/release.py
+++ b/release.py
@@ -190,6 +190,18 @@ class Tag:
     def gitname(self) -> str:
         return "v" + self.text
 
+    @property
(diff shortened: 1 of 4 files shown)


release.py
```diff
@@ -419,13 +431,30 @@ def tweak_patchlevel(
     print("done")
 
 
+def tweak_readme(tag: Tag, filename: str = "README.rst") -> None:
+    print(f"Updating {filename}...", end=" ")
+    readme = Path(filename)
+
+    # Update first line: "This is Python version 3.14.0 alpha 7"
+    # and update length of underline in second line to match.
+    lines = readme.read_text().splitlines()
```
ezio-melotti, line 440: Should this check for empty lines at the beginning of the file and strip them, or are the first two lines guaranteed to be the header?

release.py
```diff
@@ -419,13 +431,30 @@ def tweak_patchlevel(
...
+    this_is = f"This is Python version {tag.long_name}"
+    underline = f"{'=' * len(this_is)}"
```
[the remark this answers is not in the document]
hugovk: Thanks, I edited the suggestion to use double quotes.

release.py
hugovk: I don't think we don't need to, the file starts immediately with the header and has done for a long time:

https://github.com/python/cpython/blob/main/README.rst?plain=1

ezio-melotti (approved):

PR #226, opened 2025-03-22
Prompt to update the docs by version page

hugovk:
Fixes https://github.com/python/release-tools/issues/165.

Prompt to update https://www.python.org/doc/versions, which is often forgotten.

Print out the reStructuredText needed for pasting into the page.

Refactor the similar prereleases page function to also use `Tag.is_final` and simplify name.


ned-deily (approved):
LGTM, thanks!

PR #227, opened 2025-03-28
Disable uplink in OpenSSL instead of patching

zooba:


zooba:
Already tested in https://dev.azure.com/Python/cpython/_build/results?buildId=160225&view=results

PR #241, opened 2025-05-02
Add Windows release options for JIT builds

brandtbucher:
I'm not sure how to test this, or even if this is the right approach, so I'd appreciate @zooba's eyes on this. Hopefully I'm on the right track, though.

I went ahead and "future-proofed" this by adding additional options for enabling the JIT by default (not desired until 3.15 at the earliest), and also building it on free-threaded builds (not possible until 3.15 at the earliest). These options can be dropped though, if they add too much noise.

### DIFF
--- a/windows-release/azure-pipelines.yml
+++ b/windows-release/azure-pipelines.yml
@@ -41,6 +41,10 @@ parameters:
   displayName: "Signature description"
   type: string
   default: '(default)'
+- name: DoJIT
+  displayName: "Build the JIT compiler (3.14 and later)"
+  type: boolean
+  default: false
 - name: DoGPG
   displayName: "Include GPG signatures (3.13 and earlier)"
   type: boolean
@@ -49,18 +53,6 @@ parameters:
   displayName: "Include free-threaded builds (3.13 and later)"
   type: boolean
   default: true
-- name: DoJIT
-  displayName: "Build the JIT compiler (3.14 and later)"
-  type: boolean
-  default: false
-- name: DoJITEnabled
-  displayName: "Enable the JIT compiler by default (not used yet)"
-  type: boolean
-  default: false
-- name: DoJITFreethreaded
-  displayName: "Build the JIT compiler for free-threaded builds (not used yet)"
-  type: boolean
-  default: false
 - name: DoARM64
   displayName: "Publish ARM64 build (3.11 and later)"
   type: boolean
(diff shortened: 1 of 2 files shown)


windows-release/stage-build.yml
```diff
@@ -52,7 +55,12 @@ jobs:
         Platform: x86
         Configuration: Release
         _HostPython: .\python
-        ExtraOptions: ''
+        ${{ if ne(parameters.DoJIT, 'true') }}:
+          ExtraOptions: ''
+        ${{ elseif ne(parameters.DoJITEnabled, 'true') }}:
+          ExtraOptions: '--experimental-jit-off'
+        ${{ else }}:
+          ExtraOptions: '--experimental-jit'
```
zooba, line 63: I don't love the repetition of this. Two options you can try:
* add an "ExtraBuildOptions" parameter to this template and put it _unconditionally_ in all of these, then have the condition in `azure-pipelines` to choose which of these to pass.
* add a `variables` block to each job in this file with an `ExtraOptions2`, and update the build steps file to use both

Possibly some combination of the two approaches is best. The main thing I want to optimise for is only seeing the two options (i.e. `'--experimental-jit'`) once (each), since they're global across the whole matrix.

windows-release/azure-pipelines.yml
```diff
@@ -49,6 +49,18 @@ parameters:
...
+  type: boolean
+  default: false
+- name: DoJITEnabled
+  displayName: "Enable the JIT compiler by default (not used yet)"
+  type: boolean
+  default: false
+- name: DoJITFreethreaded
+  displayName: "Build the JIT compiler for free-threaded builds (not used yet)"
+  type: boolean
+  default: false
```
zooba, line 63: In terms of ordering, use descending version numbers, and put the future proofing ones right at the end. Having the version numbers in there is very important or we'll definitely get the releases wrong :)

If it makes more sense, you can make a drop down with `type: string` and `values:` like the signing certificates above. Trick here is that you can't select "none" unless it's actually in the list, and you'll need `eq()` checks (unless you can figure out a way to use the option directly and have better display text, which may be possible, but I don't know how).

zooba:
Okay, running a test build here (from a copy of this branch, not directly): https://dev.azure.com/Python/cpython/_build/results?buildId=161312&view=results

FYI @ambv, we'll have a new version-specific option to select when running a release.

zooba:
Test build was a success, so let's merge

PR #250, opened 2025-05-16
First beta: improve `SOURCE_URI` error message

hugovk:
This will help future me for 3.15.0b1.

---

Note: for 3.14.0b1, I needed to change this to `.../3.14/...` in `main` before starting the release.

And then after the release, 3.14 was in the new `3.14` branch and 3.15 had become `main`, so needed to revert it back to `.../main/...` on `main`.

### DIFF
--- a/run_release.py
+++ b/run_release.py
@@ -533,7 +533,10 @@ def check_pyspecific(db: ReleaseShelf) -> None:
         f"SOURCE_URI = 'https://github.com/python/cpython/tree/{expected_branch}/%s'"
     )
     if expected != line.strip():
-        raise ReleaseException("SOURCE_URI is incorrect")
+        raise ReleaseException(
+            f"SOURCE_URI is incorrect, expected: {expected}, got: {line.strip()} "
+            "(it needs changing before beta 1)"
+        )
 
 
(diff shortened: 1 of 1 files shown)


run_release.py
```diff
@@ -533,7 +533,10 @@ def check_pyspecific(db: ReleaseShelf) -> None:
         f"SOURCE_URI = 'https://github.com/python/cpython/tree/{expected_branch}/%s'"
     )
     if expected != line.strip():
-        raise ReleaseException("SOURCE_URI is incorrect")
+        raise ReleaseException(
+            f"SOURCE_URI is incorrect, expected: {expected}, got: {line.strip()} "
+            "(it needs changing before beta 1)"
```
ezio-melotti, line 538: I usually line up things so it's clear what parts don't match.  In this case it should be pretty easy to spot, but given that the error message is quite long and might wrap around, you might still prefer doing something like:
```suggestion
            f"SOURCE_URI is incorrect (it needs changing before beta 1):\n"
            f"expected: {expected}\n"
            f"got     : {line.strip()}"
```

sethmlarson (approved):
Thanks Hugo!

PR #254, opened 2025-06-13
Fix SBOM skip for embeddable runtime

zooba:


zooba:
FTR, this branch was for testing an updated PC/layout script that's in https://github.com/python/cpython/pull/135461, so I can go back and republish older versions. For those that pre-date SBOMs, we need to skip the script/upload entirely, but hopefully we can infer that from the source tree at the tag.

PR #255, opened 2025-06-17
Allows windows-<VER>.json files to be added to pydotorg

zooba:


add_to_pydotorg.py
```diff
@@ -271,18 +271,12 @@ def list_files(release: str) -> Generator[tuple[str, str, int, bool, str], None,
         if rfile.endswith((".asc", ".sig", ".crt", ".sigstore", ".spdx.json")):
             continue
 
-        for prefix in ("python", "Python"):
-            if rfile.startswith(prefix):
-                break
-        else:
+        prefix, _, rest = rfile.partition("-")
+
+        if prefix.lower() not in ("python-", "windows-"):
```
hugovk, line 276: I've added unit tests for this function in https://github.com/python/release-tools/pull/258.

Let's merge that first, then update this PR from ~`main`~ `master`, and it'll identify this bug -- we drop the hyphen in the `partition("-")`:

```suggestion
        if prefix.lower() not in ("python", "windows"):
```

And then also append this onto the expected file list in `test_list_files`:

```python
        (
            "windows-3.14.0b3.json",
            "Windows release manifest",
            1,
            False,
            "Install with 'py install 3.14'",
        ),
```

zooba:
Thanks for finishing up the test changes, btw

PR #8, opened 2020-12-18
Create a file to automate as much as possible the release process

pablogsal:
Behold the glorified automation project for release management: GAPRM :laughing: 

I know the file is gigantic, so feel free to not review it :)

**I have successfully used this file for the last 3 alpha releases** and I am quite happy about it so I think we can merge it and iterate over with the missing cases. I will modify it as I see that some tasks are missing as I do more kinds of releases. It also supports checkpoints, so if a task fails a new invocation will automatically resume from the last successful task!

Example output:

```
Release data:
- Branch: master
- Release tag: 3.10.0a4
- Normalized release tag: 3.10.0
- Git repo: /home/pablogsal/github/python/master
- SSH username: pablogsal
- python.org API key : .....

✅  Checking git is available
✅  Checking latexmk is available
✅  Checking make is available
✅  Checking blurb is available
✅  Checking autoconf is available
✅  Checking GPG keys
✅  Validating ssh connection to downloads.nyc1.psf.io
✅  Check buildbots are good
✅  Checking git repository is clean
✅  Checking out a temporary release branch
✅  Run blurb release
✅  Check documentation
✅  Checking git repository is clean
✅  Preparing pydoc topics
✅  Running autoconf
✅  Checking git repository is clean
✅  Bump version
✅  Checking git repository is clean
✅  Create tag
✅  Building release artifacts
✅  Test release artifacts
✅  Upload files to the PSF server
✅  Place files in the download folder
✅  Platform release managers have been notified of the release artifacts
✅  The django release object has been created
✅  Wait until all files are ready
✅  Final touches for the release
✅  Branch out new versions and prepare master branch
✅  Push new tags and branches to upstream
✅  Removing temporary release branch
✅  Add files to python.org download page
✅  Purge the CDN of python.org/downloads
✅  Modify the pre-release page
```

Example invocation:

```
python run_release.py --release 3.10.0a4 --repository=/home/pablogsal/github/python/master --auth-key=username:API_KEY_GOES_HERE
```

PR #260, opened 2025-07-11
Don't show API key in release summary

hugovk:
Right now, `run_release.py` prints a summary at the start, including the API key:

```console
❯ .venv/bin/python run_release.py --repository ../cpython --ssh-user me --release 3.14.0a0
Release data:
- Branch: main
- Release tag: 3.14.0a0
- Normalized release tag: 3.14.0
- Git repo: ../cpython
- SSH username: me
- python.org API key: me:123456789012345678901234567890123456789
- Sign with GPG: False
```

~Let's mask it:~

```console
❯ .venv/bin/python run_release.py --repository ../cpython --ssh-user me --release 3.14.0a0
Release data:
- Branch: main
- Release tag: 3.14.0a0
- Normalized release tag: 3.14.0
- Git repo: ../cpython
- SSH username: me
- python.org API key: me:************************************789
- Sign with GPG: False
```

Let's not show it:

```console
❯ .venv/bin/python run_release.py --repository ../cpython --ssh-user me --release 3.14.0a0
Release data:
- Branch: main
- Release tag: 3.14.0a0
- Normalized release tag: 3.14.0
- Git repo: ../cpython
- SSH username: me
- Sign with GPG: False
```

### DIFF
--- a/run_release.py
+++ b/run_release.py
@@ -252,7 +252,9 @@ class ReleaseDriver:
         print(f"- Normalized release tag: {release_tag.normalized()}")
         print(f"- Git repo: {self.db['git_repo']}")
         print(f"- SSH username: {self.db['ssh_user']}")
-        print(f"- python.org API key: {self.db['auth_info']}")
+        user, key = self.db["auth_info"].split(":")
+        masked = "*" * (len(key) - 4) + key[-4:]
+        print(f"- python.org API key: {user}:{masked}")
         print(f"- Sign with GPG: {self.db['sign_gpg']}")
         print()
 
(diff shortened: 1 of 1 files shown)


run_release.py
```diff
@@ -252,7 +252,9 @@ def __init__(
         print(f"- Normalized release tag: {release_tag.normalized()}")
         print(f"- Git repo: {self.db['git_repo']}")
         print(f"- SSH username: {self.db['ssh_user']}")
-        print(f"- python.org API key: {self.db['auth_info']}")
+        user, key = self.db["auth_info"].split(":")
```
sethmlarson, line 255: Is there a reason to print it at all? We can remove the whole line if it's not useful.

hugovk: It's probably not needed -- do any @python/release-managers find this useful?

It was originally added when the script was created by @pablogsal in https://github.com/python/release-tools/pull/8.

I've only once needed to see it, to double check I was using the correct API key (I'd actually got the username wrong). And I could have added a temporary print on that occasion.

AA-Turner (approved):

PR #269, opened 2025-08-14
Python 3.14 universal2 is for macOS 11 and later

hugovk:
For https://github.com/python/cpython/issues/137749.

The Python 3.14 installer is for macOS 11 and later.

---

I also added tests, which found an existing bug (tests find bugs? 🤯).

The previous `f"for macOS {'10.13' if v >= (3, 12, 6) else '10.9'} and later"` was giving us "for macOS 10.9 and later" for 3.12.10 instead of "for macOS 10.13 and later.

See: https://www.python.org/downloads/release/python-31210/

This is because we were `minor_version_tuple` only gives us `(x, y)`, not `(x, y, z)`:
```pycon
>>> v = >>> v = minor_version_tuple("3.12.10")
>>> v
(3, 12)
>>> f"for macOS {'10.13' if v >= (3, 12, 6) else '10.9'} and later"
'for macOS 10.9 and later'("3.12.10")
>>> v
(3, 12)
>>> f"for macOS {'10.13' if v >= (3, 12, 6) else '10.9'} and later"
'for macOS 10.9 and later'
```

Instead, let's add `base_version_tuple`:

```pycon
>>> v = base_version_tuple("3.12.10")
>>> v
(3, 12, 10)
>>> f"for macOS {'10.13' if v >= (3, 12, 6) else '10.9'} and later"
'for macOS 10.13 and later'
```

---

Finally, remove some redundant code for EOL Python 3.8.

### DIFF
--- a/add_to_pydotorg.py
+++ b/add_to_pydotorg.py
@@ -81,7 +81,6 @@ google_oidc_provider = "https://accounts.google.com"
 
 # Update this list when new release managers are added.
 release_to_sigstore_identity_and_oidc_issuer = {
-    "3.8": ("lukasz@langa.pl", github_oidc_provider),
     "3.9": ("lukasz@langa.pl", github_oidc_provider),
     "3.10": ("pablogsal@python.org", google_oidc_provider),
     "3.11": ("pablogsal@python.org", google_oidc_provider),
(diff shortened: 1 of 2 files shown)


ned-deily:
The basic idea LGTM, thanks. Did you go back and check the minimum supported versions? They look about right from memory :)

PR #270, opened 2025-08-14
Build Android artifacts with a reusable workflow

webknjaz:
This is a follow up for #265 that added initial implementation with
copy-pasted steps.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -41,9 +41,9 @@ permissions: {}
 
 # Set from inputs for workflow_dispatch, or set defaults to test push/PR events
 env:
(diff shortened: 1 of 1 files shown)


webknjaz:
cc @freakboy3742

mhsmith:
Thanks; I'm on holiday for the next few days, but I'll look at this when I get back.

webknjaz:
@hugovk is that necessary at this point, while in review? The branch has a few temporary things that should not be merged anyway.

hugovk:
Yes please, it's better if we can review it in the mergeable state, without the temporary things/FIXMEs.

mhsmith:
I'm not sure this is an improvement. See https://github.com/python/cpython/pull/137768#issuecomment-3215609688.

PR #271, opened 2025-08-15
Move removals above deprecations in What's New template

hugovk:
Removals are generally more important.

To match how we did it in 3.13 https://docs.python.org/dev/whatsnew/3.13.html and will do for 3.14: https://github.com/python/cpython/pull/137794.

cc @AA-Turner

### DIFF
--- a/run_release.py
+++ b/run_release.py
@@ -157,6 +157,7 @@ module_name
 -----------
 
 * TODO
+.. Add removals above alphabetically, not here at the end.
 
 
 Deprecated


ezio-melotti (approved):

run_release.py
```diff
@@ -168,6 +159,15 @@
...
 
+Deprecated
+==========
+
+* module_name:
+  TODO
+
+
+.. Add deprecations above alphabetically, not here at the end.
+
```
AA-Turner, line 170:
```suggestion


```

PR #281, opened 2025-09-18
Add Savannah for 3.16

savannahostrowski:


add_to_pydotorg.py
```diff
@@ -87,6 +87,8 @@ def run_cmd(
     "3.12": ("thomas@python.org", google_oidc_provider),
     "3.13": ("thomas@python.org", google_oidc_provider),
     "3.14": ("hugo@python.org", github_oidc_provider),
+    "3.15": ("hugo@python.org", github_oidc_provider),
+    "3.16": ("savannah@python.org", github_oidc_provider),
```
sethmlarson, line 91: Shall we add you for 3.17 as well?

PR #284, opened 2025-10-07
Add `--fast-ci` to android.py command line

mhsmith:
This is required by the changes in https://github.com/python/cpython/pull/138805.


hugovk:
Merging, this should be fine when run against the real tag.

PR #365, opened 2026-04-09
Add iOS XCframework build to release artefacts.

freakboy3742:
Adds iOS XCframework builds to the release artefacts for Python 3.15 and later.

Also adds the adjustment to the Android build to use the `Platforms/Android` script, with a compatibility shim for older releases.

See also python/cpython#140189, python/cpython#148282.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -270,7 +270,20 @@ jobs:
             mkdir -p Platforms/Android
             ln -s ../Android Platforms/Android
             ln -s ../Android/android.py Platforms/Android/__main__.py
+            echo "ADDED SYMLINK"
+          else
+            echo "NO SYMLINK"
           fi
+          ls -la Platforms/Android
+
+      - name: Check persistence
+        run: |
+          echo "PLATFORMS/ANDROID"
+          ls -la Platforms/Android
+          echo "ANDROID"
+          ls -la Android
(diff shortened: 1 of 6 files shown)


hugovk:
We can use the latest 3.15 alpha by default, similar to https://github.com/python/release-tools/pull/272 for Android.

.github/workflows/build-release.yml
hugovk:
```suggestion
GIT_COMMIT: ${{ github.event.inputs.git_commit || '55ea59e7dc35e1363b203ae4dd9cfc3a0ac0a844' }}
CPYTHON_RELEASE: ${{ github.event.inputs.cpython_release || '3.15.0a8' }}
```

freakboy3742: The downside (in the short term) is that this won't test a docs build on CI passes - but I guess it's no problem to update it again if/when we need to test a docs build.

hugovk: Yes, testing the new iOS is the priority now. 

Or we could reverse-Volkswagen it, and select all jobs if we're under test, otherwise keep the usual version selection when under release?

hugovk:
Thanks!

.github/workflows/build-release.yml
hugovk: Reverse VW mode: https://github.com/python/release-tools/pull/368

PR #368, opened 2026-04-14
Select all artifact jobs when testing

hugovk:
The "build artifacts" workflow can triggered in three ways:

* `workflow_dispatch`: when make a real release. The actual release Git remote/commit/version are passed in as inputs.
* `push` and `pull_request`: when opening PRs to test we've not broken anything. A default remote/commit/version is used.

This is a follow on from https://github.com/python/release-tools/pull/365, which changed the default for testing from 3.14 to 3.15, so we can test the new iOS job. A drawback of this is 3.15 is still in alpha, and so won't select the docs test until RC.

Instead, when we're testing, let's select _all_ jobs, so we test docs, iOS and Android. A bit like a reverse Volkswagen mode :)

Also added summaries to the build pages.

Demos:

<table>
<tr>
 <td><a href="https://github.com/hugovk/release-tools/actions/runs/24394128567">push</a>
 <td><a href="https://github.com/hugovk/release-tools/actions/runs/24394128567">workflow_dispatch</a>
<tr>
 <td>Select all
 <td>Select based on version
<tr>
 <td>image
 <td>image

</table>

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -61,6 +61,14 @@ jobs:
           echo "git_commit: $GIT_COMMIT"
           echo "cpython_release: $CPYTHON_RELEASE"
 
+          {
+            echo "| Variable        | Value                |"
+            echo "| --------------- | -------------------- |"
+            echo "| git_remote      | \`$GIT_REMOTE\`      |"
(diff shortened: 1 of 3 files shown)


freakboy3742 (approved):
All makes sense to me!

PR #373, opened 2026-05-01
Add support for new build layout in Windows free-threaded builds

zooba:


windows-release/azure-pipelines.yml
```diff
@@ -42,6 +42,10 @@ parameters:
   displayName: "Signature description"
   type: string
   default: '(default)'
+- name: Post315OutputDir
+  displayName: "Separate free-threaded outputs (3.15.0b1 and later)"
```
hugovk: We can just call it 3.15, I'm not planning on any more alphas between now and Tuesday :)

```suggestion
  displayName: "Separate free-threaded outputs (3.15 and later)"
```

zooba: Sure. It was mainly because there's a specific version check in the build now so that `PC/layout` works both pre- and post-b1. But it won't matter for any genuine builds.

zooba:
This will need https://github.com/python/cpython/pull/149218 as well to work, but feel free to merge when it all settles. I'll be offline for the weekend.

zooba:
Other PR is merged, which means this is now essential to release 3.15.

hugovk:
3.14.5 rc1 out of the way, merging this for tomorrow's 3.15 b1.

PR #378, opened 2026-05-06
windows-release/tcltk-build: Update to handle Tcl/Tk 9

zware:
- **tcltk-build: Remove Tix**
- **tcltk-build: Split {Tcl,Tk}SourceTag into version and prefix parameters**

### DIFF
--- a/windows-release/tcltk-build.yml
+++ b/windows-release/tcltk-build.yml
@@ -1,10 +1,18 @@
 parameters:
-- name: TclSourceTag
-  displayName: 'Tcl Source Tag'
+- name: TclVersion
+  displayName: 'Tcl Version'
   type: string
-- name: TkSourceTag
-  displayName: 'Tk Source Tag'
+- name: TkVersion
+  displayName: 'Tk Version'
   type: string
+- name: TclTagPrefix
+  displayName: 'Tcl source tag prefix'
+  type: string
+  default: 'tcl-'
+- name: TkTagPrefix
+  displayName: 'Tk source tag prefix'
+  type: string
+  default: 'tk-'
 - name: SigningCertificate
   displayName: "Code signing certificate"
   type: string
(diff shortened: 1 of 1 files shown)


zooba:
Apart from the CI failure, looks good to me. I guess they need some better cross-compiling support, but hopefully you can override a build variable to get it to use the Win32 build during the ARM64 one?

zware:
Yep, looking into what else needs the override now.

zooba:
Another trick we could use here is to extract the version number from the header file, that way we can go back to setting just the tags when building:

```powershell
$ver = @{}
Get-Content .\generic\tcl.h | ?{ $_ -match '#\s*define\s+TCL_([A-Z]+)_VERSION\s+(\d+).*'} | %{ $ver[$Matches[1]] = $Matches[2] }
$ver.MAJOR
$ver.MINOR
```

That doesn't help people trying to build separately (without this script), but the current change doesn't help them either.

zware:
Hmm, could we actually do that in `tcltk.props`?  I'm not too concerned about this solution being just for our purposes; once we do the update, the values in `tcltk.props`/`tcl.vcxproj`/`tk.vcxproj` will be right (for anyone sticking to the same version as us).

I think this is ready to go, but I'd rather someone else hit the big green button in this repo :)

zooba:
> Hmm, could we actually do that in `tcltk.props`?

Not so easily without adding more build-time dependencies (such as shelling out to Powershell or Python to do the command and print each variable one at a time - parsing stuff in MSBuild is a pain). It would be easier to use more wildcards and fewer hard-coded assumptions about the Tcl and Tk builds' outputs.

> I think this is ready to go, but I'd rather someone else hit the big green button in this repo

Have you been able to run test builds of it directly from this branch? It's possible, and I assume you've got the permissions, but if that's been fine then it's fine by me.

zware:
Yes; see [test build (unsigned)](https://dev.azure.com/Python/cpython/_build/results?buildId=170040&view=results),  python/cpython#149477 for CI against it (using [my fork of cpython-bin-deps](https://github.com/zware/cpython-bin-deps/tcltk)).

zooba:
Done. I'm off for the night now though, so if you start a signed build immediately you'll have to hope Hugo is about to approve it, or I'll do it tomorrow.

zware:
Thanks!

PR #383, opened 2026-05-07
Give name for iOS artifact in build-release.yml

### LINKED ISSUE #381: Give name for iOS artifact
Like the others, let's also give a `name:` in the `actions/upload-artifact` so it has a better name than just "artifact":

Image

https://github.com/python/release-tools/actions/runs/25498821336

Affects the zip filename, do this before https://github.com/python/release-tools/issues/380.

savannahostrowski:
Closes #381 

I think this is all that's needed?


hugovk (approved):
https://github.com/python/release-tools/actions/runs/25514692465?pr=383

image

Yes, looks good. Maybe lower case? I don't mind really, merge when ready!

PR #382, opened 2026-05-07
Update files for manual edit

savannahostrowski:
Based on files touched in https://github.com/python/cpython/pull/149500, this updates the files for manual edit

PR #385, opened 2026-05-07
Ask about regen-configure

zware:


StanFromIreland:
Also, magic numbers (and some Windows build things) need bumping.

StanFromIreland:
Oh just noticed https://github.com/python/release-tools/pull/382.

release.py
```diff
@@ -568,7 +568,10 @@ def bump(tag: Tag) -> None:
 
     print("Bumped revision")
     if extra_work:
-        print("configure.ac has changed; re-run autotools!")
+        if ask_question("configure.ac has changed; run 'make regen-configure'?"):
+            run_cmd(["make", "regen-configure"])
```
zware: I'm not sure what state the repository should be in at this point, so we might need to either `make -f Makefile.pre.in regen-configure` or just call `Tools/build/regen-configure.sh` directly:


```suggestion
        if ask_question("configure.ac has changed; run 'make regen-configure'?"):
            run_cmd(["make", "-f", "Makefile.pre.in", "regen-configure"])
```

```suggestion
        if ask_question("configure.ac has changed; run 'Tools/build/regen-configure.sh'?"):
            run_cmd(["Tools/build/regen-configure.sh"])
```

savannahostrowski: I'm thinking that maybe `Tools/build/regen-configure.sh` is safer? Since `bump()` is only called for fresh X.Y.0a0 alphas where the tree might never have been configured, the `make` route could be brittle here?

zware: Done!

release.py
```diff
@@ -568,7 +568,10 @@ def bump(tag: Tag) -> None:
 
     print("Bumped revision")
     if extra_work:
-        print("configure.ac has changed; re-run autotools!")
+        if ask_question("configure.ac has changed; run 'Tools/build/regen-configure.sh'?"):
```
zware:
```suggestion
        if ask_question(
            "configure.ac has changed; run 'Tools/build/regen-configure.sh'?"
        ):
```

savannahostrowski (approved):

hugovk (approved):

PR #386, opened 2026-05-07
Fix (and modify) the What's New template

StanFromIreland:
The issue on line 113 caused docs build warnings, blocking the CI. Apart from that, I removed a few things and cleaned up whitespace. I think the comments were a bit out of date, we expect contributors to write full entries now, editors only do the final touch-ups. I can revert changes if it's too much.

### DIFF
--- a/run_release.py
+++ b/run_release.py
@@ -45,42 +45,20 @@ RELEASE_REGEXP = re.compile(
 DOWNLOADS_SERVER = "downloads.nyc1.psf.io"
 DOCS_SERVER = "docs.nyc1.psf.io"
 
-WHATS_NEW_TEMPLATE = """
-****************************
+WHATS_NEW_TEMPLATE = """\
+*****************************
   What's new in Python {version}
-****************************
+*****************************
 
 :Editor: TBD
 
 .. Rules for maintenance:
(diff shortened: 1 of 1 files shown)


run_release.py
```diff
@@ -45,42 +45,20 @@
 DOWNLOADS_SERVER = "downloads.nyc1.psf.io"
 DOCS_SERVER = "docs.nyc1.psf.io"
 
-WHATS_NEW_TEMPLATE = """
-****************************
+WHATS_NEW_TEMPLATE = """\
```
hugovk: We can leave this:
```suggestion
WHATS_NEW_TEMPLATE = """
```

hugovk (commented):
Cleanups look okay.

> The issue on line 113 caused docs build warnings, blocking the CI.

Is this outdated, was it actually https://github.com/python/cpython/pull/149500#discussion_r3203790161?

image

StanFromIreland:
> Is this outdated, was it actually https://github.com/python/cpython/pull/149500#discussion_r3203790161?

No, it was the reason you had to disable branch protection ;-)

hugovk (approved):

Issue #169, opened 2024-09-25
release `bump` should either automatically update or remind to update README version to `x.y.z+`


ned-deily:
PEP 101 talks about manually updating the top-level `README.rst` file to the new version in the final steps (`bump`) of the release process as part of the post-release commit. In the past, release managers manually updated the version in the README to `3.x.y+` (i.e. `3.12.6+`) to match the PY_VERSION string automatically updated in `Include/patchlevel.h` that is displayed by sys.version and the REPL. That makes it clearer that the repo does not necessarily reflect the state of any particular release - a checkout of the release tag (`v3.12.6`) provides that - but rather something newer than a release. I think PEP 101 used to explicitly mention this but it certainly would be better to automate if possible.

hugovk:
Right now:

```console
❯ rg "This is Python version"  --sort-files
3.10/README.rst
1:This is Python version 3.10.20

3.11/README.rst
1:This is Python version 3.11.15

3.12/README.rst
1:This is Python version 3.12.13

3.13/README.rst
1:This is Python version 3.13.14

3.14/README.rst
1:This is Python version 3.14.6

3.15/README.rst
1:This is Python version 3.15.0 beta 3

main/README.rst
1:This is Python version 3.16.0 alpha 0
```

If we were to add `+` (or `+dev` for 3.15 and later), that works cleanly for fully released 3.10 to 3.14, but not so much for pre-releases; would 3.15 read "This is Python version 3.15.0 beta 3+dev"?

Some ideas:

* as above: "This is Python version 3.15.0 beta 3+dev"
* switch to the short form: "This is Python version 3.15.0b3+dev"
* ditch the micro and alpha/beta/RC: "This is Python version 3.15" or "This is Python 3.15"

ned-deily:
I think either option 1 or 2 is fine; the text doesn't have to exactly match the displayed version info in the build. What's more important is having the README be clear that this is not a tagged / released snapshot so -1 for option 3. (Same for the built docs, for that matter.)

PR #394, opened 2026-05-26
Make untagged versions PEP 440-compliant

savannahostrowski:
Related to https://github.com/python/cpython/issues/99968

### DIFF
--- a/release.py
+++ b/release.py
@@ -470,9 +470,13 @@ def tweak_patchlevel(
         "rc": "PY_RELEASE_LEVEL_GAMMA",
         "f": "PY_RELEASE_LEVEL_FINAL",
     }[tag.level]
-    new_constants = template.format(
-        tag=tag, level_def=level_def, plus=done and "+dev" or ""
-    )
+    if done:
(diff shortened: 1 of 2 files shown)


ned-deily:
While making this change, it would be good to also address the updating of the README file (#169).

hugovk:
In https://github.com/python/cpython/issues/99968#issuecomment-4527038879 you suggested targeting at least 3.16 onward, although this PR applies to all of 3.10-3.16.

If we only target 3.16, this won't have any affect until after the first 3.16 alpha in October (it's on 3.16.0a0 until then). I think it's okay to also apply this to the 3.15 beta, to help shake out any issues.

Shall we do that for a few 3.15 releases before applying it to the full set, or just go for them all now?

savannahostrowski:
@hugovk If you're cool with that, I think 3.15+ would be great. I've pushed a change for that.

@ned-deily I think I'd prefer to tackle this in a separate PR so we can keep this one scoped on the PEP 440 compliance. I've assigned myself to #169 though, so I can look at that next.

hugovk (approved):
Thanks!

PR #402, opened 2026-06-02
Quote remote release-tool command arguments

### LINKED ISSUE #397: Bug: release tooling should quote remote command arguments
`run_release.py` builds remote shell command strings for release publication.
Several values are interpolated directly into those command strings, including
release credential environment values, the release tag, and remote staging paths
derived from release-manager state.

Those values are normally trusted release-process inputs, but quoting them
consistently makes the generated commands safer and more predictable when a
value contains whitespace or shell metacharacters.

### Reproducer

Using mocked Paramiko clients:

1. Set the python.org API credential value to contain shell metacharacters, such
   as `user:key; echo unexpected`.
2. Make the fake Sigstore issuer return a token containing shell metacharacters,
   such as `token; touch /tmp/unexpected`.
3. Run the python.org upload step and capture the command sent to
   `exec_command()`.
4. The captured command includes those values unquoted in `AUTH_INFO=...` and
   `SIGSTORE_IDENTITY_TOKEN=...` assignments.

Similar mocked tests can set the release-manager SSH username to a value with a
metacharacter and observe that derived staging paths are inserted into cleanup,
copy, unzip, move, group, and permission commands without quoting.

### Expected behavior

Release-tool commands should quote environment values, release arguments, and
remote filesystem paths before sending command strings to the remote shell,
while preserving intentional glob expansion for staged artifact copies.

### LINKED ISSUE #401: Bug: Windows upload helper should quote URL-derived remote paths
`windows-release/merge-and-upload.py` converts installer metadata URLs from
`__install__.json` into remote filesystem paths. The resulting paths are then
used in SSH and SCP command arguments for directory creation, upload, group, and
mode updates.

The metadata is normally produced by the release process, but the URL-derived
path should still be treated as path data and quoted before it is passed through
remote command strings.

### Reproducer

Run this from the repository root on the current base tree. It mocks the
subprocess boundary and does not contact the upload host:

```python
import importlib.util
from pathlib import Path

script = Path("windows-release/merge-and-upload.py")
spec = importlib.util.spec_from_file_location("merge_and_upload_repro", script)
module = importlib.util.module_from_spec(spec)

try:
    spec.loader.exec_module(module)
except SystemExit:
    # The script exits when no local __install__.json files are present. The
    # upload helper functions are already defined by that point.
    pass

calls = []


def fake_run(*args, single_cmd=False):
    calls.append(args)
    return ""


module._run = fake_run
module.PLINK = "plink.exe"
module.PSCP = "pscp.exe"
module.UPLOAD_HOST = "downloads.example.org"
module.UPLOAD_USER = "release-manager"
module.NO_UPLOAD = False
module.LOCAL_INDEX = False

dest = module.url2path(
    "https://www.python.org/ftp/python/3.14.0;touch marker/"
    "python-3.14.0-amd64.exe"
)

# This matches the directory preparation code in the base upload loop.
destdir = dest.rpartition("/")[0]
module.call_ssh(f"mkdir {destdir} && chgrp downloads {destdir} && chmod a+rx {destdir}")
module.upload_ssh("python-3.14.0-amd64.exe", dest)

for call in calls:
    print(call)
```

The captured commands include the URL-derived path without shell quoting:

```text
('plink.exe', '-batch', 'release-manager@downloads.example.org', 'mkdir /srv/www.python.org/ftp/python/3.14.0;touch marker && chgrp downloads /srv/www.python.org/ftp/python/3.14.0;touch marker && chmod a+rx /srv/www.python.org/ftp/python/3.14.0;touch marker')
('pscp.exe', '-batch', 'python-3.14.0-amd64.exe', 'release-manager@downloads.example.org:/srv/www.python.org/ftp/python/3.14.0;touch marker/python-3.14.0-amd64.exe')
('plink.exe', '-batch', 'release-manager@downloads.example.org', 'chgrp downloads /srv/www.python.org/ftp/python/3.14.0;touch marker/python-3.14.0-amd64.exe && chmod g-x,o+r /srv/www.python.org/ftp/python/3.14.0;touch marker/python-3.14.0-amd64.exe')
```

The semicolon from the metadata URL remains shell syntax in the generated
directory, upload, group, and mode commands.

### Expected behavior

Remote paths derived from installer metadata URLs should be quoted as path data
before they are used in SSH command strings or SCP remote path arguments.

e-q:
## Summary

- Quote environment values and remote paths before building release upload command strings.
- Preserve intended glob expansion for staged release artifact copies.
- Quote URL-derived Windows upload paths before passing them to SSH/SCP helpers.
- Add mocked regression tests for the generated SSH and upload command strings.

Closes #397 (release tooling remote command arguments)
Closes #401 (Windows upload URL-derived remote paths)

## Testing

- Focused tox regression tests for generated remote command quoting: passed.
- `tox -q -e mypy`: passed.
- Ruff lint and whitespace checks on touched files: passed.

### DIFF
--- a/run_release.py
+++ b/run_release.py
@@ -45,6 +45,11 @@ RELEASE_REGEXP = re.compile(
 DOWNLOADS_SERVER = "downloads.nyc1.psf.io"
 DOCS_SERVER = "docs.nyc1.psf.io"
 
+
+def quote_remote_shell(value: object) -> str:
+    return shlex.quote(str(value))
+
+
 WHATS_NEW_TEMPLATE = """
 *****************************
   What's new in Python {version}
@@ -736,7 +741,7 @@ def upload_files_to_server(db: ReleaseShelf, server: str) -> None:
     ftp_client = MySFTPClient.from_transport(transport)
     assert ftp_client is not None, f"SFTP client to {server} is None"
 
-    client.exec_command(f"rm -rf {destination}")
+    client.exec_command(f"rm -rf {quote_remote_shell(destination)}")
 
     with contextlib.suppress(OSError):
         ftp_client.mkdir(str(destination))
@@ -791,29 +796,30 @@ def place_files_in_download_folder(db: ReleaseShelf) -> None:
             raise ReleaseException(channel.recv_stderr(1000))
 
     def copy_and_set_permissions(source_glob: str, destination: str) -> None:
-        execute_command(f"mkdir -p {destination}")
-        execute_command(f"cp {source_glob} {destination}")
+        quoted_destination = quote_remote_shell(destination)
+        execute_command(f"mkdir -p {quoted_destination}")
+        execute_command(f"cp {source_glob} {quoted_destination}")
         # Skip chgrp/chmod if already correct: another RM may have created
         # the directory, and only the owner can change group or permissions.
         execute_command(
-            f"find {destination} -maxdepth 0 ! -group downloads "
+            f"find {quoted_destination} -maxdepth 0 ! -group downloads "
             f"-exec chgrp downloads {{}} +"
         )
         execute_command(
-            f"find {destination} -maxdepth 0 ! -perm 775 -exec chmod 775 {{}} +"
+            f"find {quoted_destination} -maxdepth 0 ! -perm 775 -exec chmod 775 {{}} +"
         )
         execute_command(
-            f"find {destination} -maxdepth 1 -type f -user $USER ! -perm 664 "
-            f"-exec chmod 664 {{}} +"
+            f"find {quoted_destination} -maxdepth 1 -type f -user $USER "
+            f"! -perm 664 -exec chmod 664 {{}} +"
         )
 
-    copy_and_set_permissions(f"{source}/downloads/*", destination)
+    copy_and_set_permissions(f"{quote_remote_shell(source)}/downloads/*", destination)
 
     # Docs
     release_tag = db["release"]
     if release_tag.is_final or release_tag.is_release_candidate:
         copy_and_set_permissions(
-            f"{source}/docs/*",
+            f"{quote_remote_shell(source)}/docs/*",
             f"/srv/www.python.org/ftp/python/doc/{release_tag}",
         )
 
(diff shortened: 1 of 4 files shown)


sethmlarson:
@e-q Overall comment: Instead of introducing `shlex.quote()` everywhere is there value instead in changing functions to accept a list of parameters instead of a `str` and calling `shlex.join` before sending the command? Then we don't have to rely on remembering to call `shlex.quote()` everywhere.

e-q:
Gotcha, I refactored this so the SSH helpers now accept command argument lists and call `shlex.join()` at the execution boundary instead of requiring each call site to remember `shlex.quote()`.

For the two places that intentionally need remote shell glob expansion, the updated version uses static `sh -c` snippets and passes the dynamic paths as positional arguments. That keeps the wildcard behavior explicit while still joining the dynamic values centrally.

hugovk:
This now has a conflict after the merge of https://github.com/python/release-tools/pull/406.

e-q:
Rebased onto current `main` and resolved the conflict from #406 by preserving its `-maxdepth 1` / current-user permission narrowing in the list-based command form.
Local tests, mypy, and lint pass.

windows-release/merge-and-upload.py
```diff
@@ -67,6 +68,18 @@ class RunError(Exception):
     pass
 
 
+def join_remote_command(command):
```
sethmlarson: Can we add a type check here that `command` is a list or tuple. This is a safeguard as `str` is also an `Iterable[str]`.

run_release.py
```diff
@@ -45,6 +45,11 @@
 DOWNLOADS_SERVER = "downloads.nyc1.psf.io"
 DOCS_SERVER = "docs.nyc1.psf.io"
 
+
+def join_remote_command(command: list[object]) -> str:
```
sethmlarson: Same check here as above.

sethmlarson (approved):
Only had one small comment, otherwise LGTM.

e-q:
CI failures not related to this PR, addressed in #411

hugovk:
> @e-q Overall comment: Instead of introducing `shlex.quote()` everywhere is there value instead in changing functions to accept a list of parameters instead of a `str` and calling `shlex.join` before sending the command? Then we don't have to rely on remembering to call `shlex.quote()` everywhere.

Hmm, sometimes when needing to debug stuff, it's easier to have a command as a string like `"chgrp -R docs {destination}"` that can be copied and pasted with less modification, than something like `["chgrp", "-R", "docs", destination]`.

Is it possible to safely have them as strings as a starting point?

e-q:
@hugovk What do you think of a static shell template + quoted substitutions helper:
```python
  from string import Template
  from typing import LiteralString

  def join_remote_template(template: LiteralString, **values: object) -> str:
      quoted = {name: shlex.quote(str(value)) for name, value in values.items()}
      return Template(template).substitute(quoted)
```
  Then call sites can stay copy/pasteable:
```python
  execute_command("chgrp -R docs $destination", destination=destination)
  execute_command("chmod -R 775 $destination", destination=destination)
  execute_command(
      "find $destination -type f -exec chmod 664 {} \\;",
      destination=destination,
  )
  execute_command("cp $source/* $destination", source=source, destination=destination)
```
This way you can copy-paste shell-looking strings, while dynamic values are still quoted centrally. I’d use `string.Template` rather than `.format()` because these commands contain literal `{}` for `find -exec`.

hugovk:
@e-q Looks good.

@sethmlarson ok for you?

e-q:
@sethmlarson – Any preference here?

PR #403, opened 2026-06-02
Fail add_to_pydotorg on release-file create failures

### LINKED ISSUE #398: Bug: add_to_pydotorg can finish after partial release-file creation
`add_to_pydotorg.py` currently deletes the existing release-file rows before it
creates the replacement rows. If one of the later create requests fails, the
helper returns `-1`, the caller skips the failed row, and the script continues.

That can leave the release with only the subset of replacement files that were
created before the failure.

### Reproducer

Using mocked API calls:

1. Make `DELETE /downloads/release_file/?release=...` return `204`.
2. Make the first `POST /downloads/release_file/` return a created row.
3. Make the second `POST /downloads/release_file/` fail.
4. Run the publishing flow.

The script still reaches `Done - 1 files added`.

### Expected behavior

A failed release-file create should fail the publishing run. Any rows created by
that same run should be cleaned up so the command does not finish with a partial
replacement set.

e-q:
## Summary

- raise an error when creating a release-file API object fails
- clean up release-file rows created earlier in the same run if a later create
  fails
- add focused tests for create failure reporting and cleanup behavior

Closes #398

## Tests

- Focused pytest coverage for `post_object` and `create_release_files`: passed.
- Ruff lint and format checks on touched files: passed.
- Direct repository mypy check through an existing mypy environment: passed.
- Whitespace check: passed.

### DIFF
--- a/add_to_pydotorg.py
+++ b/add_to_pydotorg.py
@@ -29,7 +29,7 @@ import os
 import re
 import subprocess
 import sys
-from collections.abc import Generator
+from collections.abc import Generator, Iterable
 from functools import cache
 from os import path
 from typing import Any, NoReturn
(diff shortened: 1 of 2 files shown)


hugovk (approved):
I did the 3.14.6 release with this branch, everything uploaded fine.

https://www.python.org/downloads/release/python-3146/

sethmlarson (approved):

PR #405, opened 2026-06-02
Publish free-threaded Windows ZIP artifacts

### LINKED ISSUE #400: Bug: free-threaded Windows artifacts are skipped by add_to_pydotorg
`add_to_pydotorg.list_files()` checks that each artifact filename starts with
the target release version. It accepts names such as:

```text
python-3.14.0-amd64.exe
python-3.14.0-embed-amd64.zip
```

Free-threaded Windows ZIP artifacts use a `t` marker after the release version,
for example:

```text
python-3.14.0t-amd64.zip
```

Those names do not start with `3.14.0-` or `3.14.0.`, so the script treats them
as a different version and skips them.

### Expected behavior

The release-file publishing script should recognize free-threaded Windows ZIP
artifacts for the target release and publish them with distinct file metadata.

e-q:
## Summary

- accept the Windows free-threaded `t` version marker when listing release files
- add architecture-distinct free-threaded Windows ZIP descriptions so generated
  slugs do not collide
- add regression coverage for prerelease/final free-threaded names and generated
  metadata

Closes #400

## Tests

- Focused pytest coverage for `list_files` and free-threaded generated
  metadata: passed.
- Ruff lint and format checks on touched files: passed.
- Direct repository mypy check through an existing mypy environment: passed.
- Whitespace check: passed.

### DIFF
--- a/add_to_pydotorg.py
+++ b/add_to_pydotorg.py
@@ -129,6 +129,33 @@ def get_file_descriptions(
             ("Windows embeddable package (ARM64)", "windows", False, ""),
         ),
         (
+            rx(r"\dt-amd64\.zip$"),
+            (
+                "Windows free-threaded 64-bit embeddable package",
+                "windows",
+                False,
+                "",
+            ),
+        ),
+        (
+            rx(r"\dt-arm64\.zip$"),
+            (
+                "Windows free-threaded ARM64 embeddable package",
+                "windows",
+                False,
+                "",
+            ),
+        ),
+        (
+            rx(r"\dt-win32\.zip$"),
+            (
+                "Windows free-threaded 32-bit embeddable package",
+                "windows",
+                False,
+                "",
+            ),
+        ),
+        (
             rx(r"-arm64\.exe$"),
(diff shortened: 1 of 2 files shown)


chris-eibl:
```python
            "python-3.14.0b3t-amd64.zip",
            "Windows free-threaded 64-bit embeddable package",
``` 
Please note that the free-threaded embeddable zip files aren't produced, yet.
They'd be named like `python-3.14.6t-embed-amd64.zip`, but only the GIL variant `python-3.14.6-embed-amd64.zip` is listed on https://www.python.org/ftp/python/3.14.6/ and https://www.python.org/downloads/release/python-3146/.

`python-3.14.6t-amd64.zip` is used by the [Python install manager](https://www.python.org/downloads/latest/pymanager/), and is not the embeddable zip file. I would not list it (and definitely not as the "embeddable zip"), since we do not list its GIL equivalent `python-3.14.6-amd64.zip` either.

e-q:
Thanks, that distinction makes sense. I'll go ahead and close this.

PR #407, opened 2026-06-03
Remove redundant `check_pyspecific` task

hugovk:
https://github.com/python/cpython/pull/150850 will replace the hardcoded branch name in `SOURCE_URI` with one generated via `patchlevel`.

That means we can remove the `check_pyspecific` task from here, which is extra good because it's a disruptive one.


savannahostrowski (approved):
LGTM!

PR #415, opened 2026-06-23
Ignore failing buildbots that last ran over 14 days ago

### LINKED ISSUE #410: Ignore certain buildbots
During the 3.14.6 release:

```
The following buildbots are failing:
- wasm32-wasi Non-Debug 3.14

Check https://buildbot.python.org/all/#/release_status for more information

Do you want to continue even if these builders are failing?
Enter yes or no: yes
```

This is caused by some bad buildbot config, quickly fixed, but the builds are now part of history:

https://buildbot.python.org/all/#/builders/1918 

A couple of options to fix in release-tools:

* Ignore that specific builder
* Ignore any builders that haven't run in N days

hugovk:
Fixes https://github.com/python/release-tools/issues/410.

### DIFF
--- a/buildbotapi.py
+++ b/buildbotapi.py
@@ -62,7 +62,10 @@ class BuildBotAPI:
         }
         return all_builders
 
-    async def is_builder_failing_currently(self, builder: Builder) -> bool:
+    async def is_builder_failing_currently(self, builder: Builder) -> bool | None:
+        """Return whether the most recent build failed, or None if the
+        builder is stale (ignored because its most recent build is older
+        than STALE_BUILDER_DAYS)."""
         builds_: dict[str, Any] = await self._fetch_json(
             f"https://buildbot.python.org/all/api/v2/builds?complete__eq=true"
             f"&&builderid__eq={builder.builderid}&&order=-complete_at"
@@ -75,7 +78,7 @@ class BuildBotAPI:
 
         age_days = (time.time() - build["complete_at"]) / SECONDS_PER_DAY
         if age_days > STALE_BUILDER_DAYS:
-            return False
+            return None
         if build["results"] == 2:
             return True
         return False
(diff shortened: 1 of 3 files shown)


buildbotapi.py
```diff
@@ -1,11 +1,17 @@
...
+import time
 from dataclasses import dataclass
 from typing import Any, cast
 
 from aiohttp.client import ClientSession
 
 JSON = dict[str, Any]
 
+# Builders whose most recent build was more than this many days ago
+#  are considered inactive and ignored when checking for failures
```
savannahostrowski:
```suggestion
# are considered inactive and ignored when checking for failures
```

tests/test_buildbotapi.py
```diff
@@ -108,24 +108,36 @@ async def test_buildbotapi_stable_builders() -> None:
...
+
 @pytest.mark.asyncio
 @pytest.mark.parametrize(
-    ["json_data", "expected"],
+    ["json_data", "now", "expected"],
     [
-        ("tests/buildbotapi/success.json", False),
-        ("tests/buildbotapi/failure.json", True),
-        ("tests/buildbotapi/no-builds.json", False),
+        # Recent builds: judged on their result
```
savannahostrowski, line 121: Maybe we can add a case for the cutoff (i.e. `FAILURE_COMPLETE_AT + 14 * DAY`)

savannahostrowski (approved):
Overall, LGTM. 

Maybe we could also print out "ignoring N stale builders" or something in `check_buildbots` as well?

tests/test_buildbotapi.py
hugovk: Added in https://github.com/python/release-tools/pull/415/commits/48dcbc21c399095d23564e21b476844d723154bf.

hugovk (commented):
> Maybe we could also print out "ignoring N stale builders" or something in `check_buildbots` as well?

Added in https://github.com/python/release-tools/pull/415/changes/e63c48dec7858e3b59897710a6e6ddcf26115cc2.

Could perhaps use an enum instead of `bool | None`.

PR #416, opened 2026-06-23
Upgrade aiohttp from 3.13.3 to 3.14.1

hugovk:
Dependabot claims aiohttp needs updating at https://github.com/python/release-tools/security/dependabot from affected versions <= 3.13.3 to patched 3.13.4, but when it tries to upgrade it, also claims:

> No security update is needed as aiohttp is no longer vulnerable
> Dependabot hasn't attempted to update aiohttp as it's no longer vulnerable.

Let's just upgrade it. I ran:

```
pip-compile --generate-hashes --output-file=requirements.txt requirements.in --upgrade-package aiohttp
pip-compile --generate-hashes --output-file=dev-requirements.txt dev-requirements.in --upgrade-package aiohttp
```


ezio-melotti (approved):

ezio-melotti:
It would be good to figure out why dependabot is failing to create PRs for this (it only does for `mypy-requirements`) and fix that, but I'm the meanwhile we can keep doing it manually to ensure `aiohttp` is updated.

PR #421, opened 2026-07-10
Relabel the MSI build option and reorder

zooba:


hugovk:
Could we bunch the "3.x and earlier/later" all at the top, newest releases first?

It makes it much easier if I only need to look at the top few checkboxes and not scan the full list looking for relevant ones. (I know "mine" are already up there, but would be a good pattern to follow.)

image

zooba:
I basically see it as build features being grouped first, and then package features being grouped second, so it would annoy me a bit to disorder them like that. But I can live with it I suppose; it does make things easier to group the "things that need to be changed" above the things that don't.

hugovk (approved):
Thanks!

zooba:
The PreferredToolArchitecture change didn't help, so reverting it as it's unrelated here

PR #422, opened 2026-07-21
Automatically suffix the README version after a release

### LINKED ISSUE #169: release `bump` should either automatically update or remind to update README version to `x.y.z+`

savannahostrowski:
Closes #169 

Opted to automatically update the README (rather than remind), using option 1 from the discussion: This is Python version 3.15.0 beta 3+dev style suffixes. This happens in release.py --done, alongside the existing patchlevel.h update, so it lands in the post-release commit: + for 3.14 and earlier, +dev for 3.15+.

### DIFF
--- a/release.py
+++ b/release.py
@@ -519,6 +519,13 @@ def tweak_readme(tag: Tag, filename: str = "README.rst", done: bool = False) -> 
     lines[1] = underline
     content = "\n".join(lines)
 
+    if done:
+        # Post-release only changes the version line; the substitutions below
+        # already ran at bump time, and skipping them avoids the PEP fetch.
+        readme.write_text(content)
+        print("done")
+        return
+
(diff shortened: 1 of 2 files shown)


hugovk (approved):
Thanks!

PR #424, opened 2026-07-31
Fix Windows layout directory for both Tcl 8 and 9

hugovk:
Possible fix for https://github.com/python/cpython/pull/150197#issuecomment-5146072967 to use the `tcl8` directory for Tcl 8.6 and `tk9.0` for Tcl 9.


hugovk:
Doesn't look like we need this: https://github.com/python/cpython/pull/150197#issuecomment-5165026740

We could instead update the check to avoid the warning with v9, something for later.

